PR #6, opened 2019-09-17
Replace dfs.ItemInfo.Count with len(dfs.ItemInfo)

laurafroelich:
In Python 3.7, dfs.ItemInfo.Count raised an error that using len instead
fixes so that code reads a dfs0 file successfully.

### DIFF
--- a/pydhi/helpers.py
+++ b/pydhi/helpers.py
@@ -1,0 +1,14 @@
+def safe_length(input_list):
+    """
+    Get the length of a Python or C# list.
+
+    Usage:
+       safe_length(input_list)
+
+    input_list : Python or C# list
+
+    Return:
+        int
+           Integer giving the length of the input list.
+    """
+    return getattr(input_list, 'Count', len(input_list))
--- a/pydhi/dfs0.py
+++ b/pydhi/dfs0.py
@@ -22,7 +22,7 @@ class dfs0():
 
         dfs = DfsFileFactory.DfsGenericOpen(dfs0file)
 
-        n_items = dfs.ItemInfo.Count
+        n_items = len(dfs.ItemInfo)
         nt = dfs.FileInfo.TimeAxis.NumberOfTimeSteps
 
         names = []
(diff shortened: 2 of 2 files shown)


ecomodeller:
This requires closer investigation, since I get an error trying to use len on a C# list.


```
Python 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32

Type "help", "copyright", "credits" or "license" for more information.
>>> import clr
>>> clr.AddReference('System.Collections')
<System.Reflection.RuntimeAssembly object at 0x000002530156DC18>
>>> from System.Collections.Generic import List
>>> from System import String
>>> l = List[String]()
>>> l.Add("Foo")
>>> l.Count
1
>>> len(l)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object of type '0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]' has no len()
>>>
```

laurafroelich:
Okay, I was not aware the list could be anything but a Python list. Would we need to add some type checking to determine the list type?

j08lue:
how about
```python
def safelen(thing):
    return getattr(thing, 'Count', len(thing))
```

laurafroelich:
good idea, I like it. I tried to implement your suggestion, and it fixes the issue I was having. @ecomodeller can you check if it also works for you?

pydhi/helpers.py
```diff
@@ -0,0 +1,14 @@
+def safe_length(input_list):
+    """
+    Get the length of a Python or C# list.
+
```
j08lue: You could consider using some well-defined docstring format, [either Google or Numpy](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy).

laurafroelich: yes, I agree, would be nice to change the formatting of docstrings to one of those. I chose to follow the existing style in this repo (had a look at docstrings for other functions in the same file).

Issue #20, opened 2020-02-27
Improve handling of res1d files


ecomodeller:
- As similar as possible to other classes
- Additional functionality

rhaDHI:
I tried to fix the first point and pulled requested the changes in PR #22 . Cannot assign an issue. Maybe I don't have high enough rights within this repo?

rhaDHI:
I think there is a lot of functionality that can be added more easily if we wrap the DHI.Services.Providers.Mike1D package. It is available from our internal nuget packages and is internal open source. Then you also don't depend on your own MIKE installation. But I guess it requires some nuget package handling.

ecomodeller:
It has to be available for anyone (not only internal) in order to be used in mikeio, since this is a public repo.

4 mars 2020 kl. 15:46 skrev Rasmus Halvgaard <notifications@github.com>:

﻿

I think there is a lot of functionality that can be added more easily if we wrap the DHI.Services.Providers.Mike1D package. It is available from our internal nuget packages and is internal open source. Then you also don't depend on your own MIKE installation. But I guess it requires some nuget package handling.

rhaDHI:
@larsmichael Will this Domain Service Provider be available at some point on public nuget?

rhaDHI:
> It has to be available for anyone (not only internal) in order to be used in mikeio, since this is a public repo.

Can't we just add the dll's from the provider nuget package to the mikeio package?

ecomodeller:
Only if the license of that package permits that.

ecomodeller:
@rhaDHI Should we close this?

rhaDHI:
Yeah, we need Domain Services to be released publicly. I don't want to re-code the functionality in a new language.

larsmichael:
I am regularly advocating to make the DS GitHub repositories public, but apparently the organization is not yet mature enough to agree on this. But I will continue to advocate this.

JesperGr:
The DHI.Services.Providers.Mike1D is a fairly simple wrapper of the MIKE 1D ResultData object, implementing the DHI.Services interfaces. There is not really a lot of functionality in the DHI.Services.Providers.Mike1D - the power is only really released when the entire DHI.Services is unleashed. So, unless to use the DHI.Services for all mikeio files, I think there is no need to to just use it for MIKE 1D? 

[...]

PR #45, opened 2020-05-21
Major rewrite of res1d

maximlt:
Hi,

This PR is in a draft state and acts more as a proposal for which I kindly ask for your feedback. It contains lots of changes (sorry for the two commits only) including API breaking changes.

After noticing that the res1d example on the README was broken, I decided to try to improve the res1d module, ending up with more changes than I planned.

I started with improving the query object so that all the attributes can be set at once. It also partly validates the query. One of my goals was to be able to simply create batch queries, for instance the query `('WaterLevel', 'somebranch')` would request all the points that have water level data in *somebranch*. This goal can already be seen in the implementation of the query class. (Note that there are maybe methods in the SDK that can achieve that in a more efficient way, I didn't go through the whole documentation which is pretty hard to follow).

I have then changed the Res1D class:
* added attributes like `self.file` that are used throughout the class methods
* added a context manager to open and close the file
* added short properties to extract data
* adapted the *find* method to batch queries
* changed quite a lot of variable names (some to be a little more pythonic, which could also be done for the query object)
 
The module res1d has now a `read` function that is a wrapper around `Res1D('file').read(queries)`.

I've also added tests that all pass in the current state of the module.

I'd understand if this PR is too much at once and you ask me to revert some (or most) of the changes made here. Yet I'm sure there are a few good ideas thrown in there that you could keep in the future.


rhaDHI:
Nice work Maxime :-)
I PR'ed some formatting to your 'maximlt:improve-res1d' branch. 
Please also review the missing res1d functionality issue here: https://github.com/DHI/mikeio/issues/20
There will be a new query class making this stuff easier in the next Mike1D.ResultDataAccess release.
Thanks!

rhaDHI:
@maximlt If you can also update the repo ReadMe with examples, that would be awesome.

maximlt:
Thanks for your review @rhaDHI ! I didn't pay too much attention to formatting as I considered this PR as a proposal, I appreciate your input though!

I had already started to work on this PR and missed #20, I see there's more to come and that's good news.

As for this PR, if you're fine with it I guess you could merge it :)

I've got one comment (for a future PR maybe) about the names of the `QueryData` attributes. If there's no good reason for them to follow the camel case style (like `VariableType`), they'd rather be named in a more pythonic way (like `variable_type`).

I'd also like to share a video of a conference talk given by Raymond Hettinger where he explains how to write a Python adaptor library for an SDK written in another language, I found it very instructive.
https://www.youtube.com/watch?v=wf-BqAjZb8M

rhaDHI:
OK cool thanks. Will check out the video.
I agree with the naming convention. The guy who wrote that class followed the C# standard instead of python. There is a new query class in the next Mike1D release so not sure we even need it. But feel free to change it for consistency.

maximlt:
If you don't need it it's fine, it's just that we had a use case and we may use this version of `res1d` for the time being.

mikeio/res1d.py
```diff
@@ -1,141 +1,274 @@
...
 import os.path
 import pandas as pd
 
 clr.AddReference("DHI.Mike1D.ResultDataAccess")
-from DHI.Mike1D.ResultDataAccess import ResultData
+from DHI.Mike1D.ResultDataAccess import ResultData  # noqa
 
 clr.AddReference("DHI.Mike1D.Generic")
-from DHI.Mike1D.Generic import Connection
+from DHI.Mike1D.Generic import Connection  # noqa
```
rhaDHI, line 12: # noqa. What does this mean?

mikeio/res1d.py
```diff
@@ -1,141 +1,274 @@
...
+            raise TypeError("chainage must be either None or a number.")
+        if bn is None and c is not None:
+            raise ValueError("chainage cannot be set if branch_name is None.")
+
+    @property
+    def variable_type(self):
+        return self._variable_type
+
+    @property
+    def branch_name(self):
```
rhaDHI, line 257: Nice. Properties is a better way.

tests/test_res1d.py
```diff
@@ -1,28 +1,108 @@
...
+    ts = read(file, [q1, q2])
+    assert ts.shape == (110, 2)
+    ts_max = ts.max()
+    assert pytest.approx(round(ts_max[0], 3)) == 197.046
+    assert pytest.approx(round(ts_max[1], 3)) == 0.761
+
+
+def test_read_reach(file):
+    q_reach = QueryData("WaterLevel", "118l1")
+    ts = read(file, [q_reach])
```
rhaDHI, line 86: We should create a read function with a single query as input to avoid brackets

mikeio/res1d.py
maximlt: That's how to tell Python linters to skip that line, see https://stackoverflow.com/questions/45346575/what-does-noqa-mean-in-python-comments

I've added that comment here since my linter in VSCode was complaining about this import as it couldn't resolve it (due to the fact that this import is resolved at runtime). I assume other linters will have the same error in other editors, if you want to add a linter check in your CI in the future it's likely you'll have to add `# noqa` at the end of each one of these imports (there might be a global setting though).

maximlt: I can remove it if you like!

mikeio/res1d.py
maximlt: Yep I've chosen to use properties here so that the attributes have to be validated only once at the initialization.

mikeio/res1d.py
rhaDHI: No need. It has a purpose.

tests/test_res1d.py
maximlt: Yes the read function could check the type of the query. I'd leave that for another PR if possible.

rhaDHI: Added it here: https://github.com/maximlt/mikeio/pull/2

rhaDHI:
If you are happy with these changes and functionality, let us know and we can merge into master and publish a new package version to make it available for everyone.

maximlt:
Ok I'll let you know when I think it's ready after some more tests.

maximlt:
Hi @rhaDHI,

We've further checked the new method and found out a bug that occurs when a res1d file contains a branch that is split in different reaches sharing the same name (but different ID internally). That case is not covered by the res1d file included in the test data (I cannot include the file we've tested).

In the process of fixing this bug I realized that the method `_find_points` was just doing too much at the same time and that it would be difficult to extend it in the future. There are now two more methods. The first one validates each query against the data contained in the file. The second one expands the queries, for instance `[QueryData("WaterLevel", "branch1")]` becomes `[QueryData("WaterLevel", "branch1", 0), ..., QueryData("WaterLevel", "branch1", 100)]`. While it's working fine, on the downside it's a little more constitutionally expensive since each method may read the file (to validate the queries, to build the queries, to find the data points).

I've also added a decorator to each public property of `Res1D` to make sure that they're not accessed when the file is not opened.

maximlt:
I have some more suggestions for improving the module.

* I would like to be able to use the Res1D class as follows:
  ```python
  # The recommended way, i.e. with a context manager
  with Res1D.open("file.res1d") as r1d:  # r1d being an instance of Res1D, not the .file attribute
      rn = r1d.reach_names
  # Do something with rn here...
  
  # Not recommended but still available in the public API as it maybe convenient sometimes.
  r1d = Res1D.open("file.res1d")
  rn = r1d.reach_names
  r1d.close()
  ```

  Would it fit with the general API of mikeio? I wouldn't like the res1d module to be too different from the other ones (if it's not already the case, I'm not familiar with the other modules). If yes, I could implement it.

* Also note that I may have been a little too strict with adding the `_not_closed` decorator to each public property of Res1D. After all once the data is retrieved (e.g. the reach names), there's no need for the file handle to be still open to access that data, it's already there. Let me know if you want to keep this behavior or want me to revert it. If so, it'll still require to handle when a property is accessed before the file is opened but that should be straightforward.

* *reach* and *branch* are used interchangeably withing the code and the API, I'd rather stick to one and I'm +1 on *branch* as in Mike Hydro.

* We'd like to access to more variable types than *WaterLevel* and *Discharge* (and *Pollutant* although I haven't tried it):
  1. *Left (and Right) outflow to Mike 21* which is a result of a 1D/2D simulation run with Mike Flood
  2. From the additional results file such as *Discharge in structure*, *Froude*, etc.

  I'd like to implement 1. in priority. Would you like me to pursue in the PR or open another one after this one is merged?

* It'd be a good idea to replace the existing `res1d` file in the test data by one that would cover the case I've mentioned in my previous post. In the event that I implement 1., a file with the associated data would be required for testing. Do you happen to have such a file (or several files if required) available?

rhaDHI:
Great reflections.

I think @ecomodeller can advise on how the API should fit to the rest of mikeio. 

@JesperGr might have an example file and some opinions on the rest of your questions.

JesperGr:
A few comments: 

Naming it Res1D is not really fully covering, since it does open res11, PRF, CRF, XRF, TRF and more, basically all files from MIKE 1D, MOUSE and MIKE 11. Someone also once wrote a "bridge" for one of our competitors output data to load into MIKE 1D ResultData object.

There is a meaning in a not-closed property on the Res1D object, because you may load the data in the file in a lazy manner, if you are only interested in some time steps of the file. But it is not required on each property in the Res1D object - whenever the header is loaded, all static data (reaches, nodes, catchments) are loaded.

Reach and branch: MIKE 11 and MIKE HYDRO calls it branches. MIKE URBAN calls it links and canals. The generic term in MIKE 1D is reaches. So, I would definitely go for reaches!

Variable types: There is in the MIKE 1D ResultData object no preference for water level and discharge, all types of data is stored in the same manner: Froude, velocities, AD components, M21 exchange flow etc. It is just a matter of requesting data with another ID. The only real difference is the Discharge in structures, which are defined on a single point, and not the entire reach.

And then do exploit the opportunity to extent the MIKE 1D ResultDataQuery class to make things easier in the python code... 

When quering data on a reach, we usually split it in three versions:
- Retrieve data for all grid points in the reach: GetReachValues(reachName). That supports reaches split in two due to junctions.
- Retrieve data for all grid points within a span of the reach: GetReachValues(reachName, startChainage, endChainage). That also supports reaches split in two due to junctions.
- Retrieve data for a single grid point in the reach: GetReachValues(reachName, chainage). Either find the closest grid point or apply interpolation.

ecomodeller:
Please go ahead and adapt the API to be able to use a context manager. The other modules of mikeio needs this functionality as well, and the API will be modified anyway.

maximlt:
Hi!

I messed up a little bit while committing my last work to this PR, I believe it's fixed but let me know if you see something wrong in the git history.

Thanks for your comments!

> Naming it Res1D is not really fully covering, since it does open res11, PRF, CRF, XRF, TRF and more,  basically all files from MIKE 1D, MOUSE and MIKE 11. Someone also once wrote a "bridge" for one > of our competitors output data to load into MIKE 1D ResultData object.

I wasn't aware of that! I've decided to keep the `Res1D` name, in the future there could be a base class for sharing common utilities for reading all these 1Dish files.

> There is a meaning in a not-closed property on the Res1D object, because you may load the data in the file in a lazy manner, if you are only interested in some time steps of the file. But it is not required on each property in the Res1D object - whenever the header is loaded, all static data (reaches, nodes, catchments) are loaded.

I came to realize that even after `file.Dispose()` some of the content of the file is still available.

> Reach and branch: MIKE 11 and MIKE HYDRO call it branches. MIKE URBAN calls it links and canals. The generic term in MIKE 1D is reaches. So, I would definitely go for reaches!

Done.

> Variable types: There is in the MIKE 1D ResultData object no preference for water level and discharge, all types of data is stored in the same manner: Froude, velocities, AD components, M21 exchange flow etc. It is just a matter of requesting data with another ID. The only real difference is the Discharge in structures, which are defined on a single point, and not the entire reach.

I've added the ability to query Left and Right outflow data. I believe it'd be really easy now to add similar data types like Froude or velocities. It's a little more tricky for DischargeInStructure, I won't add that in this PR.

> And then do exploit the opportunity to extent the MIKE 1D ResultDataQuery class to make things easier in the python code...

For sure if there's is some existing code in the SDK that is able to perform those queries, it'd be nice to use it directly instead of repeating the logic here in Python.


I haven't managed to implement the context manager as I described it above. Instead I've made the following changes inspired from the built-in `open` function:
* I've added the `open` function (could be renamed) that returns a `Res1D` instance and that can be used as follows:
  ```python
  from mikeio import res1d
  with res1d.open("file.res1d") as file_reader:
      datatypes = file_reader.data_types
  # do something with 'datatypes'
  ```
* `Res1D` now opens the file on initialization. The class could be extended in the future to accept a `mode` attribute that would indicate whether the file is on *read* or *write* mode, and modify the initialization accordingly. Now it's implicitly on *read*. `Res1D` is also a context manager.

Note that in most cases the `read` function should be enough to extract data from the file. Using `open` is for fine grained processing of the results and more convenient access to the file itself through the `file` attribute.

Regarding the tests:
* I see that you've implemented some CI testing! 👍 It's failing though now, I guess that's because the SDK isn't installed in the process.
* I've tested this PR locally with a *res1d* file containing Left and Right Outflow data. I can't share this file. It'd be nice if in the future you could add a test file including that data, as long as another one including additional result data.

I guess we're now getting closer to the end of this PR, thanks again for your comments and guidance. And feel free to change/revert whatever you think that doesn't fit your needs.

rhaDHI:
@ecomodeller I don't have write access to merge PRs in this repo. But I think we are good to go?

ecomodeller:
Thank you @maximlt for this contribution and @rhaDHI for reviewing.

PR #49, opened 2020-06-11
Minor clean up of dfs0 read

rhaDHI:


rhaDHI:
Why use FileNotFoundError?

We should change it in res1d as well then?: https://github.com/DHI/mikeio/blob/fda66bd083301447548eea5425f04fc14b7e05a4/mikeio/res1d.py#L107

ecomodeller:
To me it makes sense to use `FileNotFoundError` when trying to read a file, but the file is not found :-) 

`FileExistsError` would be used if you want to avoid overwriting existing files.

> Raised when trying to create a file or directory which **already** exists

https://docs.python.org/3/library/exceptions.html#FileExistsError

ecomodeller:
Might as well change this as well to FileNotFoundError https://github.com/rhaDHI/mikeio/blob/b257062759010a4e87fc1d0bba130b22d4080417/mikeio/dfs0.py#L119

rhaDHI:
OK. I think the other functions should be cleaned up as well. And I guess all warnings should be exceptions? But maybe in another PR.

PR #52, opened 2020-06-17
Read all items Including Z coordinate

ecomodeller:


jsmariegaard:
Should we consider making the check:

if **item == 0** and items[item].name == "Z coordinate":

instead of 

if items[item].name == "Z coordinate":

? 

In MIKE FM model dfsu-3d-outputs, the z item is always the first item. But as a dfs-manipulator-guy I may decide to make a new item with z-values in the cell centers and call the item "Z coordinate". Not very likely, but I guess the extra check does no harm?

jsmariegaard:
Otherwise it looks super good :-)

PR #62, opened 2020-08-10
add some features to mesh.py

mohm-dhi:
It is nice to be able to have contour plot in mikeio. I used matplotlib.tri to do so. It is just working on fully triangular mesh. It needs to have data on the nodes so I wrote a function to convert cell-centered data to note-centered data. 
using plot()
Plot

using plot2()  we can rename it to more meaningful name!
Plot2

### DIFF
--- a/mikeio/mesh.py
+++ b/mikeio/mesh.py
@@ -112,7 +112,7 @@ class Mesh:
 
         return mp
 
-    def plot(self, cmap=None, z=None, label=None):
+    def plot(self, cmap=None, z=None, label=None, vmin=None, vmax=None):
         """
         Plot mesh elements
 
@@ -137,6 +137,12 @@ class Mesh:
             if label is None:
                 label = "Bathymetry (m)"
 
+        if vmin is None:
+            vmin = z.min()
+
+        if vmax is None:
+            vmax = z.max()
+
         patches = []
 
         for j in range(ne):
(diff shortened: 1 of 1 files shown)


mohm-dhi:
@ecomodeller and @jsmariegaard : if you have any comments on this please let me know?

ecomodeller:
I don't think plot2 is a very descriptive name, could we add this as an option, with a suitable name, to the exisisting plot method instead?

mohm-dhi:
Hi @ecomodeller 
I agree, and I mentioned this in my first comment too. We can name it something like "contour". It can also be an option to the already existing plot method by adding a input argument such as "do_contour = True". If you agree with this, I can do the necessary changes and send another pull request.

ecomodeller:
I suggest to add input argument  "do_contour = True".

jsmariegaard:
Good idea, @mohm-dhi , with a contour plot. I agree with @ecomodeller on the input argument. Maybe add a check for quad meshes? It would be nice to have something which works with all meshes :-)

mohm-dhi:
Thanks, @jsmariegaard ! If the domain has quad meshes, first they are converted to triangular meshes (by connecting the opposite corners) and the rest would be similar to the case where we have all triangular mesh. I already applied this in the new version that I have. 

I will do this things and will send another pull request once I am done.

ecomodeller:
Would you like to update this PR to fit with the recent changes to the Dfsu,Mesh?

mohm-dhi:
Yes, sure. I was busy with moving to a new place! I haven’t had a chance to update it. I will look into it this week.

PR #67, opened 2020-08-25
Dfsu and mesh major revision

ecomodeller:


jsmariegaard:
Some remaining issues:
* all internal indexes should be zeros based (element_ids, node_ids, element_table) - subtract 1 when reading from dfsu, add 1 when writing back to file
* plotting should be enabled for dfsu files also 
* mesh write() function should use class properties instead of source 
* mesh write() function should be able to save only specific elements
* 3d stuff should be moved to geometry class 
* can we simplify class structure so we don't need two abstract classes? 
* or should we another abstract class such that Unstructured2D <- Unstrctured3D <- ... 
* vertical average
* extract z-level from 3d file
* re-projection ( @mohm-dhi )
* save 2d file from 3d file (e.g. layer or mesh)
* write mesh from dfsu (is it already possible?)

jsmariegaard:
We also need the following:
* 3d support for find_closest_element_index - if user supplies x,y for a 3d file, we should return all elements in the relevant column (or top element?)
* using tree for searching in elements
* a lot of new tests :-)

ecomodeller:
If I try to read 3700 elements, then why do I get a 4180?

And saving a 2d from a 3d does not work yet... (not even as a 3d with single layer)
image

jsmariegaard:
> If I try to read 3700 elements, then why do I get a 4180?
> 
> And saving a 2d from a 3d does not work yet... (not even as a 3d with single layer)

Right, I did not test that yet. It should work on 2d files however. Could you please test that?

jsmariegaard:
Maybe you also saw that on my above list of remaining issues :-)

* save 2d file from 3d file (e.g. layer or mesh)

jsmariegaard:
For some reason my lazy loading of e.g. dfs.element_coordinates is no longer lazy. It apparently calculated the element_coordinates on initialization which is super annoying because it makes it slow :-( 

Any ideas on why this is? It worked before...

jsmariegaard:
By the way, you wrote "If I try to read 3700 elements, then why do I get a 4180?". 

That is not correct. You are getting 3700, but with 4180 nodes (that is what ds is showing - the length of the first item which is zn). I think the extraction is correct.

jsmariegaard:
The dfsu file now contains the following properties and functions:

# Basic Properties
n_nodes       
node_coordinates
n_elements
element_coordinates
codes
valid_codes
is_2d
projection_string
is_geo
is_tri_only
type_as_string


# Less-essential properties
node_ids   
element_ids
boundary_codes
max_nodes_per_element
is_local_coordinates


# Properties "3d" - geometries with vertical axis
n_layers
n_sigma_layers
n_z_layers
top_elements
n_layers_per_column
bottom_elements
layer_ids


# Properties - dfsu files only
deletevalue
n_items
items
start_time
n_timesteps
timestep
end_time


# Advanced properties
element_table
geometry2d
e2_e3_table
elem2d_ids


# Basic functions
write(outfilename, elements=None)
plot(z=None, elements=None, label=None, cmap=None)          
find_nearest_element(x, y, z=None, layer=None)
get_element_area()


# Dfsu functions
read(items=None, time_steps=None, elements=None)  
to_mesh(outfilename)

  
# Mesh functions
set_z(z)
set_codes(codes)
plot_boundary_nodes(boundary_names)


# Advanced functions
find_nearest_profile_elements(x, y)
get_layer_elements(n)
to_shapely()
elements_to_geometry(elements, node_layers='all')    
to_2d_geometry()   

  
# Hidden/internal functions
_read_header(filename)
_read_mesh_header(filename)
_read_dfsu_header(filename)
_get_element_coords()
_set_nodes_from_source(source)
_set_elements_from_source(source)    
_get_element_coords_from_source()
_get_element_table_from_dotnet()
_element_table_to_dotnet(elem_table=None)
_set_nodes(node_coordinates, codes=None, node_ids=None, projection_string=None)
_set_elements(element_table, element_ids=None, geometry_type=None)
_reindex()
_get_top_elements_from_coordinates(ec=None)
_get_nodes_for_elements(elements, node_layers = 'all')
_get_nodes_and_table_for_elements(elements, node_layers = 'all')
_get_2d_to_3d_association()
_geometry_to_mesh(outfilename, geometry)
_get_element_table_for_elements(elements)
_find_n_nearest_elements(x, y, z=None, n=1, layer=None)
_to_polygons(geometry=None)


# Obsolete
get_node_coords(code=None)

jsmariegaard:
The following is still missing (and should in my opinion be done in a later pull request):

* spatial interpolation (top priority in my opinion) - should be fast - so maybe we need a vectorized wrapper for Jesper Grooss MeshUtil
* use search tree
* allow find_nearest_element to take coordinates fun([x,y,z]) rather than fun(x,y,z)
* allow find_nearest_element to optionally take an element id instead of coordinates (find_neighbors)
* allow find_nearest_element to return n nearest
* vertical average
* element volumes (similar to get_area)
* re-use of matplotlib polygons doing several plots faster
* extract z-level from 3d file
* use of dynamic zn values (how should we do this in a good way) (e.g. for extracting profile)
* extrapolate profile to bottom and surface optionally
* re-set zn values of 3d files (set new surface... find_sigma_depth(), ...)
* plotting of vertical slices (make new pseudo x-axis (relative in meters to start) and plot xz-plane with polygons)
* print item names in __rep__ similar to dataset

More to put on the wish list?

ecomodeller:
> The following is still missing (and should in my opinion be done in a later pull request):
> 
> * spatial interpolation (top priority in my opinion) - should be fast - so maybe we need a vectorized wrapper for Jesper Grooss MeshUtil
> * use search tree
> * allow find_nearest_element to take coordinates fun([x,y,z]) rather than fun(x,y,z)
> * allow find_nearest_element to optionally take an element id instead of coordinates (find_neighbors)
> * allow find_nearest_element to return n nearest
> * vertical average
> * element volumes (similar to get_area)
> * re-use of matplotlib polygons doing several plots faster
> * extract z-level from 3d file
> * use of dynamic zn values (how should we do this in a good way) (e.g. for extracting profile)
> * extrapolate profile to bottom and surface optionally
> * re-set zn values of 3d files (set new surface... find_sigma_depth(), ...)
> * plotting of vertical slices (make new pseudo x-axis (relative in meters to start) and plot xz-plane with polygons)
> * print item names in **rep** similar to dataset
> 
> More to put on the wish list?

I added items to the repr, (I thought I had done that already...)

PR #72, opened 2020-09-03
Add contour option to dfsu.plot

mohm-dhi:
I added do_contour as an option to dfsu.plot() method. 

It works for both triangular and quadrilateral and flexible mesh. It needs data at the nodes so I wrote a function that produce node based data from the cell centered data using the Pseudo Laplacian interpolation, similar to MeshNodeInterpolation (https://github.com/DHI/DHI.Mesh). Whenever  Pseudo Laplacian fails I used inverse distance for interpolation. 

image

Please feel free to optimize the code if it is required.

### DIFF
--- a/mikeio/dfsu.py
+++ b/mikeio/dfsu.py
@@ -918,7 +918,52 @@ class _UnstructuredGeometry:
 
         return mp
 
-    def plot(self, z=None, elements=None, label=None, cmap=None):
+    def get_node_centered_data(self, data, extra_allowed=True):
+        nc = self.get_node_coords()
+        ec = self._get_element_coords()
+        elements = [list(self._element_table[i]) for i in range(len(list(self._element_table)))]
+        if self.is_tri_only:
+            elements = np.asarray(elements)
+        else:  # if it is a mix of tri and quad
+            tmp_elements = elements.copy()
+            for el , item in enumerate(tmp_elements):
+                if len(item)==4:
+                    elements.pop(el)
+                    elements.insert(el,item[:3])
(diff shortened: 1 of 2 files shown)


jsmariegaard:
Cool stuff. I will have a closer look tomorrow and test it out. 

Speaking of tests could you maybe add one or two to test_dfsu.py?

jsmariegaard:
Hi Mohammed, I really like the stuff you have done! It also gave me some further ideas - see https://github.com/DHI/mikeio/tree/mohm-dhi-master (I am not sure they ended up the right place - I am still very much a github newbie). 

* There was in issue with plotting from 3d files - I semi-fixed this. 
* The refinement took very long time in my opinion so I made that optional.
* I think it could be nice with a plot_type and plot_mesh arguments so I added those. 
* I added n_levels as an argument so you can control the number of contour levels. 
* I cannot make the patch plot not plot the mesh- do you know how? 
* Some renaming also
* extracted a function

mohm-dhi:
Thanks Jesper! 
I like the changes you made. 
Regarding your question to not to plot mesh in the patch. you should change this lines of code to 
        p = PatchCollection(
            patches, cmap=cmap, edgecolor="lightgray", alpha=0.8, linewidths=0.3
        )

to 

        p = PatchCollection(
            patches, cmap=cmap, edgecolor="face", alpha=1 , linewidths=1
        )

or 

        p = PatchCollection(
            patches, cmap=cmap, edgecolor="face", alpha=None , linewidths=None
        )

jsmariegaard:
Hi Mohammed, I've opened a new pull request #75  with same content

PR #76, opened 2020-09-07
Handle dfsu contour_lines

mohm-dhi:
this line of code is needed 
ax.tripcolor(triang, zn, edgecolors='face', vmin=vmin, vmax=vmax, cmap=cmap)
otherwise if the vmin and vmax do not cover the entire zn range it plots white area. 

with: 
image

without would be like this! 
image

### DIFF
--- a/mikeio/dfsu.py
+++ b/mikeio/dfsu.py
@@ -1082,11 +1082,12 @@ class _UnstructuredGeometry:
             else:
                 # must be contourf plot then
                 levels = np.linspace(vmin, vmax, n_levels)
+                ax.tripcolor(triang, zn, edgecolors='face', vmin=vmin, vmax=vmax, cmap=cmap)
                 tr_fig = ax.tricontourf(triang, zn, levels=levels, cmap=cmap)
-                # if plot_type == 'contour_lines':
-                #     ax.tricontour(triang, zn, levels=levels,
-                #             colors=['0.5'],
-                #             linewidths=[0.5])
+                if plot_type == 'contour_lines':
+                    ax.tricontour(triang, zn, levels=levels,
+                            colors=['0.5'],
+                            linewidths=[0.5])
             
             plt.colorbar(tr_fig, label=label)
 


jsmariegaard:
Good stuff! Henrik and I talked about this at the office today. We couldn't fix it right away and wanted to move before the BLUE cafe tomorrow... :-)

jsmariegaard:
Hi 

I have a few more things that I would like to include in this PR: 

1) I re-factored too hard when I extracted the method _create_tri_only_element_table() - the "data" part is missing
2) _create_tri_only_element_table() produces wrong element centers - they should instead be calculated based on the new element_table - this can be done relatively easy by adopting _get_element_coords() to use nc and element_table as arguments instead of self
3) I think we should change "plot_mesh" to "show_mesh" to avoid this: mesh.plot(plot_mesh=True)
4) we should have quad (or even better mixed) mesh in testdata

I am working on most of these.

mohm-dhi:
Regarding item #4 you can use this dfsu file that I generated for testing purpose. 
image


[FakeLake.zip](https://github.com/DHI/mikeio/files/5190296/FakeLake.zip)

jsmariegaard:
I am working with this on https://github.com/DHI/mikeio/tree/mohm-dhi-master - but busy today and tomorrow. I will continue on Monday.

mikeio/dfsu.py
```diff
@@ -1082,11 +1082,12 @@ def plot(self, z=None, elements=None, label=None, cmap=None, vmin=None, vmax=Non
             else:
                 # must be contourf plot then
                 levels = np.linspace(vmin, vmax, n_levels)
+                ax.tripcolor(triang, zn, edgecolors='face', vmin=vmin, vmax=vmax, cmap=cmap)
```
jsmariegaard, line 1085: @mohm-dhi, why do you want to add this line before tricontourf()? What difference does it make?

mikeio/dfsu.py
```diff
@@ -1082,11 +1082,12 @@ def plot(self, z=None, elements=None, label=None, cmap=None, vmin=None, vmax=Non
...
                 tr_fig = ax.tricontourf(triang, zn, levels=levels, cmap=cmap)
-                # if plot_type == 'contour_lines':
-                #     ax.tricontour(triang, zn, levels=levels,
-                #             colors=['0.5'],
-                #             linewidths=[0.5])
+                if plot_type == 'contour_lines':
```
jsmariegaard, line 1087: @mohm-dhi, do you know how we could make a non-filled contour plot (which I guess is relevant expectation if a user requests a contour plot)

image

mikeio/dfsu.py
mohm-dhi: as I explained here https://github.com/DHI/mikeio/pull/76#issue-481622110, without this line when the vmin and vmax are not in the range of the data it creates white area. Please see the figures.

mikeio/dfsu.py
mohm-dhi: well, tricontour creates the lines and tricontourf make it filled contour. So if we want to have just the lines we should not have tricontourf

jsmariegaard:
Overall for the plotting, I think it would be great if we have these options for dfsu-files: 

plot_type:

1. mesh_only
2. patch (aka box)
3. shaded
4. contourf (filled contour)
5. contour (lines with labels)
6. outline_only

The types 2, 3 and 4 should have the option: show_mesh=True/False. I am almost done with this.

mikeio/dfsu.py
jsmariegaard: right - sorry

jsmariegaard:
I think it is pretty much complete now: https://github.com/DHI/mikeio/tree/mohm-dhi-master - please let me know what you think

Issue #77, opened 2020-09-08
Support for PFS files


ecomodeller:
Read pfs file into suitable python data structure. A dictionary is simple, but a bit clumsy to use for deeply nested structures. Any suggestions?
```
from mikeio import Pfs

d = Pfs("foo.m21fm")

sources = d["FemEngineHD"]["HYDRODYNAMIC_MODULE"]["SOURCES"]

n_sources = sources["number_of_sources"]

for i in range(n_sources):
  key = f"SOURCE_{i+1}"
  source = sources[key]
  name = source["name"]
```

Or perhaps:
```
m = Pfs("foo.m21fm")
m["FemEngineHD/HYDRODYNAMIC_MODULE/SOURCES"]
```

Or:

```
p = Pfs("foo.m21fm")
p.FemEngineHD.HYDRODYNAMIC_MODULE.SOURCES
```

One additional idea is to strip out unused modules during read

```
p = Pfs("foo.m21fm", strip=True)

p['PARTICLE_TRACKING_MODULE']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'PARTICLE_TRACKING_MODULE'
```

jsmariegaard:
I would really like to have this feature!

For MIKE FM setups you have only one root element e.g. FemEngineHD and beneath that typically just a few sections: 

    [DOMAIN]
    [TIME]
    [MODULE_SELECTION]
    [HYDRODYNAMIC_MODULE]
    [DATA_ASSIMILATION_MODULE]

It would be convenient if I could access these simply by 

    pfs.domain
    pfs.time
    pfs.module_selection
    pfs.HD
    pfs.DA

Furthermore, I would like to get simple dataframes from typical lists that occur in the pfs files, e.g.

    df = pfs.HD.get_sources()
    df = pfs.HD.get_outputs(included_only=True)

This could be used for plotting all sources and outputs on a map... 

I would also like to get lists of input files: 

    filenames = pfs.get_input_files()

I would be great if you could easily search - similar to what you can with EUMTypes in mikeio :-)

ecomodeller:
![pfs](https://user-images.githubusercontent.com/614215/92734215-9e8eb880-f378-11ea-9759-03a99f85d7fa.gif)

Elsafti:
[...]

PR #78, opened 2020-09-09
Support for reading PFS

ecomodeller:
Cover basic parts of #77


ecomodeller:
Must not forget to add possibility of writing pfs files. This is relevant for automating simulations. Better think through this scenario before merging this PR.

Issue #73, opened 2020-09-04
Optional dependencies


rhaDHI:
It doesn't seem right to include matplotlib and its dependencies as requirements in this lightweight file read/write package. Any idea how to manage optional dependencies? The same applies to xarray in issue #36 . Also jupyter is needed to view notebooks in the repo but should not be installed as part of the default package.

ecomodeller:
Optional dependancies can be added in the extra_requires part of setup.py
https://github.com/DHI/mikeio/blob/fd8fb73076b65456ed9fc3b18ce0219ffbc4c63c/setup.py#L10

At the moment two categories are defined:
* dev (used for development, we should add the jupyter and xarray to this list)
* test (used for testing on GitHub)

After cloning the repo, you can install like this (also mentioned in the [Contributions guidelines](https://github.com/DHI/mikeio/blob/master/CONTRIBUTING.md))

`pip install -e .[dev]`

PR #80, opened 2020-09-10
Make matplotlib optional

ecomodeller:
This should be closing #73


rhaDHI (commented):
So a default package pip install will not install matplotlib, and if you run some of the built-in mikeio plot functions it will complain about installing matplotlib. So you have to install matplotlib manually if needed. Is that the behavior you are going for?

rhaDHI (commented):
What about xarray?

rhaDHI (approved):

PR #81, opened 2020-09-11
Dfs0,1,2 repr

ecomodeller:


jsmariegaard:
would it make sense, to have a describe() method for all classes and then let __repr__ call that?

ecomodeller:
> would it make sense, to have a describe() method for all classes and then let **repr** call that?

I don't know, what is the added benefit of: `dfsu.describe()` over `repr(dfsu)`?

jsmariegaard:
> > would it make sense, to have a describe() method for all classes and then let **repr** call that?
> 
> I don't know, what is the added benefit of: `dfsu.describe()` over `repr(dfsu)`?

Nah, I guess there aren't any... just similar pattern as pandas (but that's because they have something else for __repr__). I can see that the current behaviour corresponds to xarray. Which is good. Let's keep it like that :-)

PR #85, opened 2020-09-15
More dfsu and mesh plotting

jsmariegaard:
contains #76 and more


jsmariegaard:
image

mikeio/dfsu.py
```diff
@@ -1033,74 +1062,110 @@ def plot(
...
-        if vmin is None:
+        if plot_data and vmin is None:            
             vmin = z.min()
-
-        if vmax is None:
+        if plot_data and vmax is None:
             vmax = z.max()
 
+        # set levels
+        if plot_type[:7] == 'contour':
```
ecomodeller, line 1083:
```suggestion
        if 'contour' in plot_type:
```

tests/test_dfsu_plot.py
```diff
@@ -38,7 +33,21 @@ def test_plot_3d():
 def test_plot_dfsu_contour():
     filename = os.path.join("tests", "testdata", "HD2D.dfsu")
     dfs = Dfsu(filename)
-    dfs.plot(plot_type="contour", n_levels=5)
+    dfs.plot(plot_type="contour", levels=5)
```
ecomodeller, line 36: Add a test where levels is a list

mikeio/dfsu.py
```diff
@@ -1110,42 +1175,76 @@ def plot(
...
         for el, item in enumerate(tmp_elmnt_nodes):
             if len(item) == 4:
                 elem_table.pop(el)
-                elem_table.insert(el, item[:3])
-                elem_table.append([item[i] for i in [2, 3, 0]])
-                ec = np.append(ec, ec[el].reshape(1, -1), axis=0)
+                elem_table.insert(el,item[:3])
+                elem_table.append([item[i] for i in [2,3,0]])
+                data = np.append(data, data[el])
+                ec = np.append(ec, ec[el].reshape(1, -1), axis=0)    # THIS IS NOT CORRECT!
```
ecomodeller, line 1245: Interesting comment

ecomodeller (approved):

mikeio/dfsu.py
bhlevca: I don't understand why spaces have been removed after comas. The  standard in Python is to put space after each language element. The only difference from that standarf that I know of is when defining default parameters in a functions' header when  the '=' operator is linked to the variable name and its default value. Ex: variable='10' instead of variable = '10'

tests/test_dfsu_plot.py
jsmariegaard: yes, I thought I did that - will do

mikeio/dfsu.py
jsmariegaard: me neither - will add back in

jsmariegaard: yes, forgot all about that calculation of ec - I will fix that

tests/test_dfsu_plot.py
jsmariegaard: done

mikeio/dfsu.py
jsmariegaard: done

jsmariegaard:
Would be good with dfsu/mesh plotting on frontpage readme...

jsmariegaard:
Now adding short example for readme and more references to notebooks

mikeio/dfsu.py
ecomodeller: > I don't understand why spaces have been removed after comas. The standard in Python is to put space after each language element. The only difference from that standard that I know of is when defining default parameters in a functions' header when the '=' operator is linked to the variable name and its default value. Ex: variable='10' instead of variable = '10'

We rely on [Black](https://black.readthedocs.io/en/stable/) for formatting code, which can be used to auto-format on save, so you don't have to remember the rules of a style guide.

mohm-dhi:
It looks great. Thanks! I think we should open a new project to do re_projection as it is very important and it was brought up by some of the clients as well.

mikeio/dfsu.py
bhlevca: I use PyCharm for Python work and it complained about formatting not conforming to PEP 8.  PyCharm also verifies and rearranges the code. If you use PyCharm, Black would not be necessary. 

If you look at the lines I mentioned, there are inconsistencies, some commas have spaces after them some don't. This means that either Black failed or has not been used for the file. 
Also, I noticed that Black is in "beta"

mohm-dhi:
@jsmariegaard 
We also need to add the capability to plot vertical profiles as well. I already have a code that can do that but with new changes in dfsu file I got lost how to add this feature to the plot. 

This is what I plot: 

Figure_2

jsmariegaard:
Let's add that and more to this new branch https://github.com/DHI/mikeio/tree/more_dfsu - maybe start in a new method _plot_transect() or something?

mohm-dhi:
I agree, I may work on it next week. 

One thing that you can edit in the current dfsu.plot method is to add the layer_number argument to plot dfsu3D file with sigma and/or z layer in a different layer.

jsmariegaard:
Sorry, forgot to answer this. You can already plot a specific layer:

elem_layer = dfs.get_layer_elements(-1) 
dat = dfs.read(items='my_item', time_steps=0, elements=elem_layer)[0]
dfs.plot(dat, elements=elem_layer)

mohm-dhi:
Thanks Jesper !

mohm-dhi:
Thanks Jesper !
I wrote my code to return the elements and index for each layer. Something is wrong in the .get_layer_elements because it does not return back the right elements 


@jsmariegaard

jsmariegaard:
@mohm-dhi thanks for making me aware - I thought it worked but I can see now that it didn't. Check out https://github.com/DHI/mikeio/commit/b9892db17e299f475e4a4f46f68749bdf3c154b5 in the [more dfsu branch](https://github.com/DHI/mikeio/tree/more_dfsu)

Can you please verify that it is working correct now?

jsmariegaard:
@mohm-dhi any progress on the nice vertical profile plots you showed above? Are they ready to enter the Dfsu class?

mohm-dhi:
Hi Jesper,
Sorry for the late response. I was slammed by lots of work and was busy to finish that part. I hope to find some time next week.

jsmariegaard:
Yeah I know the feeling :-) 
Looking forward to it.

PR #89, opened 2020-09-16
Option to read single precision. Closes #40

### LINKED ISSUE #40: Single vs double precision
Is it correct to convert all data to double precision
https://github.com/DHI/mikeio/blob/b6a3272e507d0c7594c9f6077dc3f9d3061458f3/mikeio/dfsu.py#L73

Or should it be `dtype=np.float32` (i.e. single precision)?

ecomodeller:
Most changes, introduced by auto-format

For the actual change, scroll down to line 1400

### DIFF
--- a/mikeio/dfsu.py
+++ b/mikeio/dfsu.py
@@ -981,7 +981,7 @@ class _UnstructuredGeometry:
         label=None,
         cmap=None,
         vmin=None,
-        vmax=None,        
+        vmax=None,
         levels=10,
         n_refinements=0,
         show_mesh=True,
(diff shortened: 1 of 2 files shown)


jsmariegaard (approved):
Looks good!

ecomodeller:
Do you think it makes sense to have the dtype argument on the init method or should it be in the read method?

Is is possible to create dfsu with double precision?

I realized that there is a similar argument on the dfs0.create method
https://github.com/DHI/mikeio/blob/dad2f2cb9bb9e5a4b7b17c6725efd8a28cf41bf0/mikeio/dfs0.py#L264

PR #91, opened 2020-09-18
Temporal interpolation of Dataset

ecomodeller:


jsmariegaard (changes requested):
Good stuff. 

I don't understand the notebook - far too short for my taste. A compact example is good but this is crazy! 

For the interp function itself, I would like to have the option to provide a list of datetimes myself - such that both could be achieved: 

     >>> dsi = ds.interp(dt=1800)

or 
     >>> dsi = ds.interp(times=otherds.timesteps)

Say, that I want to time interp my model output to match the (non-equidistant) times of an observation time series. This can, of course, come later if you would like to close this pull request fast. 

Finally, are you sure about the method name? We will probably soon add some spatial interpolation too... maybe better to call the method time_interp() or similar? In pandas and xarray I guess this is called resample() https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#resampling

ecomodeller:
1. Added some content to notebook
2. Added possibiltiy to interpolate to time axis from another dataset, dataframe etc.
3. Renamed to `time_interp`  pandas uses a two step process .resample(freq).interpolate()

jsmariegaard:
@ecomodeller do you reckon that time_interp() is better than interp_time() ? - I agree that interp is better than resample in this context. That I mentioned the first in my review was pretty random.

mikeio/dutil.py
```diff
@@ -301,6 +302,86 @@ def isel(self, idx, axis=1):
...
+                intime,
+                dataitem,
+                axis=0,
+                kind=kind,
+                copy=copy,
+                bounds_error=bounds_error,
+                fill_value=fill_value,
+                assume_sorted=assume_sorted,
+            )
+            t_out = f_out(outtime)
```
jsmariegaard, line 380: rename t_out ? maybe dataitem_out or dataitem_interp

ecomodeller:
> @ecomodeller do you reckon that time_interp() is better than interp_time() ? - I agree that interp is better than resample in this context. That I mentioned the first in my review was pretty random.

I let the 🎲 decide...

I will change it to `interp_time`

jsmariegaard:
I like that you pass arguments to interp1d() but I would have preferred "method" over "kind" and "bounds_error" will not be easy for me to remember (would probably have preferred "extrapolate=True/False"). I certainly see the point in re-using the same argument names but https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.interpolate.html which also wraps scipy interp1d still call it "method". Hmm? What do you think?

jsmariegaard:
Forgot "copy"...

Are we sure we want to expose copy argument? When should I use it?

jsmariegaard (approved):

rhaDHI:
@ecomodeller Please try to avoid abbreviations to improve readability :-) Use 'interpolate' instead of 'interp'.

ecomodeller:
Ok, ok...

Do you think the default case should be to raise an exception in case of extrapolation, or silently fill with NaN?

rhaDHI:
Scipy's default is to fill with NaNs, so follow that?

ecomodeller:
> Scipy's default is to fill with NaNs, so follow that?

No, the default behaviour is to raise an exception. But I find that a bit over-protective. So let's fill with NaN, what could possibly go wrong.

jsmariegaard:
> @ecomodeller Please try to avoid abbreviations to improve readability :-) Use 'interpolate' instead of 'interp'.

I prefer "interp" as this is what numpy, scipy and xarray use. See e.g. https://numpy.org/doc/stable/reference/generated/numpy.interp.html and http://xarray.pydata.org/en/stable/interpolation.html 

I generally agree with you @rhaDHI on explicit rarther than implicit :-) But we are not "interpolating time" anyway and interpolate_data_in_time() is too long.

Issue #94, opened 2020-09-23
Mesh().plot() for global mesh not working as expected


Hendrik1987:
**Describe the bug**
After pip install of the latest development version on 2020-09-23 I experience problems with msh.plot() for a global mesh.

**To Reproduce**
The following code

```
from mikeio import Mesh
meshfilename = r"a_global_mesh_file.mesh"
msh = Mesh(meshfilename)
msh.plot()
```
throws the following error

```
Traceback (most recent call last):

  File "<ipython-input-2-1c07926ee84f>", line 1, in <module>
    msh.plot()

  File "C:\Users\hewr\Anaconda3\lib\site-packages\mikeio\dfsu.py", line 1227, in plot
    ax.plot(*domain.exterior.xy, color=out_col, linewidth=1.2)

AttributeError: 'MultiPolygon' object has no attribute 'exterior'
```

and produces the plot in below screenshot.

**Expected behavior**
A plot like the first one in https://github.com/DHI/mikeio/blob/master/notebooks/Mesh.ipynb is the expected result.

This line on the other hand does not throw an error, but the plot is still the same.

`msh.plot(show_outline=False)`

The following, however, works as expected

```
msh.plot_boundary_nodes()
mp = msh.to_shapely()
```

**Screenshots**
![image](https://user-images.githubusercontent.com/29595985/94047978-1cb87980-fdd3-11ea-95be-64e6b9e6a124.png)

**System information:**
 - Python version 3.7.6.final.0
 - MIKE version 2020

ecomodeller:
Could you tell me the output of this code for the global mesh?
```
mp = msh.to_shapely()
domain = mp.buffer(0)
type(domain)
```

The expected outcome is `<class 'shapely.geometry.polygon.Polygon'>`

Hendrik1987:
The output of
```
mp = msh.to_shapely()
domain = mp.buffer(0)
type(domain)
```
is
`shapely.geometry.multipolygon.MultiPolygon`

(using Spyder 4)

jsmariegaard:
@Hendrik1987 also reported that shapely returns an error: "Multipolygon object has no attribute 'exterior'" upon running this line:

    ax.plot(*domain.exterior.xy, color=out_col, linewidth=1.2)

He gets the same error with domain.interiors... 

His shapely version is 1.6.4 - same as mine.

jsmariegaard:
He successfully used shapely to find if points where in domain - with these lines:

# only offshore points
[...]

PR #97, opened 2020-09-26
Fix plotting of global mesh

jsmariegaard:
fix aspect ratio, try-catch on shapely, use geometry instead of self for shapely


jsmariegaard:
Fixing #94

ecomodeller:
I suggest to change `print` to `warnings.warn` like https://github.com/DHI/mikeio/blob/e9d6999bf110e489ee9e27454193be753aebc65e/mikeio/dfs0.py#L325

PR #101, opened 2020-10-01
Relative time axis

ecomodeller:


mikeio/dfs0.py
```diff
@@ -73,7 +73,11 @@ def _read_header(self):
         self._items = get_item_info(dfs, list(range(self._n_items)))
 
         # Read time
-        self._start_time = from_dotnet_datetime(dfs.FileInfo.TimeAxis.StartDateTime)
+        try:
```
jsmariegaard, line 76: maybe better to check FileInfo.TimeAxis.TimeAxisType? (or at least easier to understand)

PR #103, opened 2020-10-02
Common read method for dfs1 and dfs2

ecomodeller:


mikeio/dfs1.py
```diff
@@ -11,21 +11,15 @@
...
 
-from .dutil import Dataset, find_item, get_item_info, get_valid_items_and_timesteps
-from .dotnet import (
-    to_numpy,
-    to_dotnet_float_array,
-    to_dotnet_datetime,
-    from_dotnet_datetime,
-)
+from .dotnet import to_dotnet_float_array
 from .eum import ItemInfo
```
rhaDHI, line 15: Remove unused imports ItemInfo + timedelta

rhaDHI (commented):
factory and builder variables can be replaced by self._factory and self._builder ?

jsmariegaard:
Lot's of good stuff here! 

May suggest to change the class name AbstractDfs to something else, maybe Dfs123, GriddedDfs or similar. The current name indicates that it's the base class of dfs classes which it is not.

PR #105, opened 2020-10-02
Use custom exceptions in dfs1

rhaDHI:
Custom exceptions is good practice in python.
Should be used to streamline exception messages and error handling across the code.
Exceptions must also be tested.
Only implemented in dfs1 in this PR.

### DIFF
--- a/mikeio/custom_exceptions.py
+++ b/mikeio/custom_exceptions.py
@@ -1,0 +1,4 @@
+class FileDoesNotExist(Exception):
+    def __init__(self, filename=""):
+        self.message = f"File {filename} does not exist."
+        super().__init__(self.message)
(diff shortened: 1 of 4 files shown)


mikeio/custom_exceptions.py
```diff
@@ -0,0 +1,10 @@
+class FileDoesNotExist(Exception):
```
ecomodeller, line 1: Seems redundant, better to use [FileNotFoundError](https://docs.python.org/3/library/exceptions.html#FileNotFoundError)

mikeio/custom_exceptions.py
```diff
@@ -0,0 +1,10 @@
...
+    def __init__(self, filename=""):
+        super().__init__(f"File {filename} does not exist.")
+
+
+class DataDimensionMismatch(ValueError):
```
ecomodeller, line 6: Useful info!

ecomodeller (approved):

mikeio/custom_exceptions.py
rhaDHI: Right. So I will change to FileNotFoundError in that case 👍

PR #111, opened 2020-10-07
Use new Mike1D Res1D queries and expose C# objects

rhaDHI:
Update Mike1D + DFS dll's to 19.0.0-beta1 (latest DHI-internal nuget package)
WARNING: Please review dll changes and remove obsolete files.

Exposes raw C# objects (specifically "query") as python properties to access all features of underlying res1d classes. Isn't that nice to be able to get those objects

WARNING:  contains breaking changes:
Does not yet support two nice features from previous versions: Separate QueryData and contextmanager way of handling open/closed file. I added a few examples including docstrings from C# to see how to use. 

Should we wrap all query methods, e.g. GetReachValues ? A lot of work and must be updated if MIKE changes API.
Otherwise the type conversion from C# to python is handled outside by the user 👎 

    data = res1d.query.GetReachValues("104l1", 34.4131, "WaterLevel")
    data = np.fromiter(data, np.float64)
 
So needs more work but illustrates the new Mike1D ResultDataQuery methods.


rhaDHI:
@maximlt What do you think? Should I add 1) the context manager way of handling the file again and 2) the querydata.

maximlt:
Hi @rhaDHI,

Nice to see you're still developing `mikeio` :+1:, it's great to finally have access to your underlying C# code.

FYI what I think about these changes don't matter so much now as I've switched job now and I'm not gonna use Mike anymore. But of course for my former colleagues it'd be nice not to break too much their code which I believe would require the context manager and a QueryData object that allows to retrieve all the values of a quantity on a branch.

ecomodeller (commented):
Add more examples on how to use it

rhaDHI:
I added more examples in test and partly updated the notebook.
Still need to add chainage to the name properly.
Still need to update readme.

rhaDHI:
@ecomodeller Did I add enough examples? Not sure there are anyone else who can review this.
Please note that this PR introduces breaking changes in res1d. 
It also updates mike bin dll's to latest mike version beta-1 from internal nuget server. Not 100% how many dll's we actually need. Some can probably be removed.

ecomodeller (approved):

PR #114, opened 2020-10-17
2d interpolation for dfsu

jsmariegaard:


mikeio/dfsu.py
```diff
@@ -488,6 +499,119 @@ def _get_element_coords(self):
...
+        x = np.linspace(x0, x1, nx)
+        y1 = y0 + dy*(ny-1)
+        y = np.linspace(y0, y1, ny)
+        xv, yv = np.meshgrid(x, y)
+        xy = np.empty((nx*ny, 2))
+        xy[:,0] = xv.reshape(-1)
+        xy[:,1] = yv.reshape(-1)
+        return self.get_2d_interpolant(xy, n_nearest)
+
+    def interp2d(self, data, interpolant2d):
```
ecomodeller, line 598: interp2d could be extracted into a new class

jsmariegaard: good idea

mikeio/dfsu.py
```diff
@@ -488,6 +499,119 @@ def _get_element_coords(self):
...
+                nx = int(np.ceil(xr/yr))*ny
+            dx = xr/(nx-1)
+            dy = yr/(ny-1)
+        else:
+            if dx is None:
+                nx, ny = shape
+                dx = xr/(nx-1)
+                dy = yr/(ny-1)
+            if shape is None:
+                dy = dx
```
ecomodeller, line 557: Allow dx to be either None, float or a two-item list or tuple to get different dx and dy

jsmariegaard: Done

mikeio/dfsu.py
jsmariegaard: I have now extracted functionality into a new file - but could not really see the benefit of making it a class - maybe you can?

ecomodeller: No.

jsmariegaard: Okay - but I'll guess we will keep it in a separate file for now, right?

PR #127, opened 2020-11-11
Read write data value type

laurafroelich:
* Add input argument to specify DataValueType, with same default value as previously set automatically,, (http://doc.mikepoweredbydhi.help/webhelp/2019/DHI_DFS/html/T_DHI_Generic_MikeZero_DFS_DataValueType.htm#!) to ItemInfo
* Update reading and writing of header to take DataValueType into account

### DIFF
--- a/tests/test_dfs2.py
+++ b/tests/test_dfs2.py
@@ -6,7 +6,6 @@ from mikeio.dfs2 import Dfs2
 from mikeio.eum import EUMType, ItemInfo, EUMUnit
 from mikeio.custom_exceptions import (
     DataDimensionMismatch,
-    InvalidDataType,
     ItemNumbersError,
 )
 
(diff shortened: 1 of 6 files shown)


tests/test_dfs2.py
```diff
@@ -362,3 +364,45 @@ def test_reproject_defaults(tmpdir):
     assert dfs.start_time == newdfs.start_time
     assert dfs.projection_string != newdfs.projection_string
 
+
+def test_write_static_item(tmpdir):
```
ecomodeller, line 368:
```suggestion
def test_write_accumulated_datatype(tmpdir):
```

ecomodeller (commented):
Please add a test method to verify that the default datatype is instantaneous.

tests/test_dfs2.py
laurafroelich: good idea, thanks!

PR #129, opened 2020-11-12
Raster

ecomodeller:
In response to:

> ... I want an asc file (or tif, or other common non-mike gis raster formats). ... I use MIKE Zero to convert dfsu to dfs2, then I use the MIKE Zero tool called mike2grd to convert the dfs2 to asc. Because they're several large dfsu files, this takes quite some time. So I am wondering if mikeio could be used to automate this whole process, from dfsu to asc.


jsmariegaard (commented):
Following our conversation yesterday on node- vs center-based grids, I will make some changes to the Grid2D class now :-)

jsmariegaard (commented):
Following our conversation yesterday on node- vs center-based grids, I will make some changes to the Grid2D class now :-)

jsmariegaard:
Are you happy now @ecomodeller ? Do we need more?

PR #135, opened 2020-11-17
Reading PFS with comments

### LINKED ISSUE #125: PFS files with comments
When I pass my pfs file (with comments) I get the following error: 

    ParserError: while parsing a flow sequence
    in "<unicode string>", line 29563, column 18
    did not find expected ',' or ']'
    in "<unicode string>", line 29564, column 9

The mentioned line contains this comment: 

    include = 1			// 0: inactive, 1: active used for DA, 2: active but only for validation

But it seems that several text strings in the comments can cause the parser to fail. 

I suggest a pre-processor which strips all "//" and following characters in the same line, before passing the pfs-file. 

**System information:**
 - Python 3.8
 - mikeio 0.6

jsmariegaard:
Should fix #125

### DIFF
--- a/mikeio/pfs.py
+++ b/mikeio/pfs.py
@@ -117,8 +117,8 @@ class Pfs:
             if s[-1] == "]":
                 s = s.replace("]", ":")
 
-        s = re.sub(re.compile("//.*?\n"), "", s)
-        s = s.replace("|", "")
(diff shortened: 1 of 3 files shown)


ecomodeller (approved):

jsmariegaard:
closes #125

Issue #138, opened 2020-11-18
Cannot import mikeio on offline computer without MIKE software


laurafroelich:
**Describe the bug**
We have downloaded mikeio 0.6.0 and its dependencies from pypi. On computers with internet and preinstalled MIKE software, including MIKE Zero, we can install and import mikeio without a problem. However, on a computer without internet access and no preinstalled MIKE software, we get the error:

    mikeio\__init__.py:22: in <module>
    from .dfs0 import Dfs0
    mikeio\dfs0.py:8: in <module>
    from DHI.Generic.MikeZero import eumQuantity
    E   ModuleNotFoundError: No module named 'DHI.Generic.MikeZero'; 'DHI.Generic' is not a package

**To Reproduce**
On a computer with internet, download mikeio and dependencies:

    pip download mikeio -d mikeio_offline

Copy the folder mikeio_offline to a computer without internet and without preinstalled MIKE software, and run the following:

    pip install  mikeio --no-index --find-links=mikeio_offline

Then start python and type

    import mikeio

**Expected behavior**
Import of mikeio with no error messages

**Screenshots**

**System information:**
 - Python version: 3.7.7
 - MIKE version: 0.6.0

ecomodeller:
Can you import any other .NET dll?

rhaDHI:
It is the first dll import it meets. So not necessarily related to that specific dll but the whole bin folder that it cannot find.
The test pipeline runs in a new environment without MikeZero, and internet access should not matter?

rhaDHI:
Maybe try mikeio 0.5.x before we updated MIKE dll's?

ecomodeller:
> It is the first dll import it meets. So not necessarily related to that specific dll but the whole bin folder that it cannot find.
> The test pipeline runs in a new environment without MikeZero, and internet access should not matter?

Exactly, the test pipeline (Actions) runs on a server without MikeZero, which should eliminate any *"works on my machine"* problems.

laurafroelich:
We initially tried with version 0.5.3 and updated to 0.6.0, hoping that would solve the issue

laurafroelich:
I agree, internet access should not matter. However, it is the only difference between the machine we are trying to use mikeio on, and the test machines that run pipelines, that I could think of. There might be other differences of course, that have not occurred to me.

rhaDHI:
32 bit problems?

rhaDHI:
Maybe try to import a dll using pythonnet on the machine?

[...]

PR #139, opened 2020-11-19
New MIKE dlls v19.0.0

rhaDHI:
From NuGet packages DHI.Mike1D.ResultDataAccess (includes dfs and eum) and DHI.Mike1D.CrossSectionModule (for PFS).


jsmariegaard:
The Gold version should be ready (at least the installers are since last week). So why use RC2?

rhaDHI:
There was a lot of unneeded files and I was cleaning them up anyway trying to solve https://github.com/DHI/mikeio/issues/138 . But let's update this PR with the actual release files. Didn't know they were that close 👍

jsmariegaard:
The release is out https://www.mikepoweredbydhi.com/mike-2021

ecomodeller:
> There was a lot of unneeded files and I was cleaning them up anyway trying to solve #138 . But let's update this PR with the actual release files. Didn't know they were that close 👍 This was the newest on the internal nuget feed.

@rhaDHI Will you update with the final release?

rhaDHI:
Yep. The release is also on public nuget now. The public nuget dll's v19.0.0 are the exact same as the internally released 19.0.0-rc2 that included. So you can go ahead and merge.

PR #143, opened 2020-11-26
Hande delete values

ecomodeller:


mikeio/generic.py
```diff
@@ -69,35 +70,38 @@ def _clone(infilename, outfilename):
...
+    infilename: str, outfilename: str, offset: float = 0.0,
+    factor: float = 1.0, items: Union[None, List[str], List[int]] = None,
 ):
     """Apply scaling to any dfs file
 
         Parameters
         ----------
 
-        infilename: str
+        infilename:
```
jsmariegaard, line 81: why remove the type from the docstring? Isn't this useful for the user also when you use type hints?

mikeio/generic.py
```diff
@@ -193,9 +203,11 @@ def diff(infilename_a, infilename_b, outfilename):
 
             itemdata_a = dfs_i_a.ReadItemTimeStep(item + 1, timestep)
             d_a = to_numpy(itemdata_a.Data)
+            d_a[d_a == deletevalue] = np.nan
```
rhaDHI, line 206: Use:
nan_indices = np.isclose(data, self._dfs.FileInfo.DeleteValueFloat, atol=1e-36)
data[nan_indices] = np.nan

??

mikeio/generic.py
ecomodeller: Maybe, I figured it was redundant, but there is no reason to remove them as long as it is consistent with the type hint.

jsmariegaard: I think it's still relevant for the user - but maybe I missed something

mikeio/generic.py
jsmariegaard: is this necessary? can we decide on a "atol" which will always work? What is fastest if the files are huge ?

PR #144, opened 2020-11-27
Add plot_vertical_profile to dfsu.py and created one test to check it

mohm-dhi:
Well I made some changes to the dfsu code. It was changed a lot from the last time that I worked on it so it was a little time consuming for me to understand it correctly. I hope you can help to merge this new method with the plot method. maybe we can add an argument or just an if statement that for vertical profile use the new method. I am open to any suggestions but as I am very busy with my other projects, it would be hard to find time to work on it. 

Thanks,


jsmariegaard (approved):
Great work @mohm-dhi ! I have made some changes to synchronize arguments with the other plotting function and added a notebook. What do you think?

mohm-dhi:
Great ! Thanks Jesper. It looks very good to me.

mikeio/dfsu.py
```diff
@@ -1149,6 +1125,101 @@ def get_node_centered_data(self, data, extrapolate=True):
...
 
+    def _Get_2DVertical_elements(self):
+        if (self._type == DfsuFileType.DfsuVerticalProfileSigmaZ) or (
+            self._type == DfsuFileType.DfsuVerticalProfileSigma
+        ):
+            elements = [
+                list(self._source.ElementTable[i])
+                for i in range(len(list(self._source.ElementTable)))
+            ]
+            return np.asarray(elements) - 1
```
ecomodeller, line 1136: Perhaps add comment here why we need to subtract 1

mikeio/dfsu.py
```diff
@@ -1149,6 +1125,101 @@ def get_node_centered_data(self, data, extrapolate=True):
...
+
+    def plot_vertical_profile(
```
ecomodeller, line 1138: Would it make sense to create a new specialized class only for vertical slices, and rename this method to plot?

jsmariegaard:
@ecomodeller, I think this is ready for merge - what about you? I am not authorized to merge so I think we need your help.

mikeio/dfsu.py
jsmariegaard: Yes, good idea

jsmariegaard: Then no matter what dfsu file you open, you have a relevant plot() function - we should then change a few things in the original dfsu.plot() e.g. rename first argument to values

PR #145, opened 2020-11-28
Dataset: create_time, create_empty_dataset, improved _init_

jsmariegaard:
Make it easier to create new datasets and hence also new files. It should be easier to create dfsu from mesh file. It is now with this PR: 

    dfs = Dfsu('mymesh.mesh')
    ds = Dataset(dfs.n_elements, time='2018-1-1', items=2) 
    dfs.write('empty_1step_2items.dfsu',ds)


mikeio/dataset.py
```diff
@@ -543,6 +565,95 @@ def to_dataframe(self, unit_in_name=False):
     def _ipython_key_completions_(self):
         return [x.name for x in self.items]
 
+    @staticmethod
+    def create_empty_data(n_items=1, n_timesteps=1, n_elements=None, shape=None):
```
ecomodeller, line 569:
```suggestion
    def empty(n_items=1, n_timesteps=1, n_elements=None, shape=None):
```

tests/test_dataset.py
```diff
@@ -705,8 +705,55 @@ def test_properties_dfsu():
...
+
+    t = Dataset.create_time("2018", dt=7200, end_time="2019")
+    assert len(t) == (365 * 12 + 1)
+
+
+def test_init():
+    n_elements = 45
+    data = n_elements
+    time = "2018-01-01"
+    items = [ItemInfo("Foo")]
```
ecomodeller, line 753:
```suggestion
    items = ["Foo"]
```

mikeio/dataset.py
jsmariegaard: empty_data() ? It's not returning an empty Dataset,,,

PR #148, opened 2020-12-02
Progress bar

ecomodeller:
I think we should add progress bars to all potentially long-running methods.

This is how it looks like:
progress


jsmariegaard:
Ooooh, fancy. I like it

PR #153, opened 2020-12-16
Dfsu plot outline only

mohm-dhi:
Hi, 
I was trying to create some graphs (about 30 graphs) and my mesh file size is about 3M. the code to plot the mesh outline takes too long and when I debugged it I found out that it is  the _get_boundary_polylines() method that takes too long so I decided to create a property to just return all polylines. it is very fast and it cab be used in the plot outline_only case. 

Thanks, 
Mohammad


jsmariegaard:
Hi @mohm-dhi , thanks a lot for this PR. Sorry for being so slow in getting back to you. My computer crashed badly on Thursday morning - got it back Friday with a new hard drive - spend a lot of time installing stuff and then this morning it was completely dead again. So I am writing this from an old spare PC :-) 

Anyways, I miss a huge mesh to test this - could you please send me one (mail is fine)? 

And then a couple of questions:
* is _get_all_boundary_polylines() supposed to replace _get_boundary_polylines() ? 
* if so would it be possible to optimize the existing method instead? 
* does the new method support disjoint domains? and domains with one or more holes? 
* could you please explain your algorithm in words (or add a few explaining comments in the code)? 

 Thanks

jsmariegaard:
And there is something wrong with outlines for 3d dfsu - try to run this notebook https://github.com/DHI/mikeio/blob/master/notebooks/Dfsu%20and%20Mesh%20-%20Plotting.ipynb 

image

mohm-dhi:
Hi @jsmariegaard 
to answer your questions. 
is _get_all_boundary_polylines() supposed to replace _get_boundary_polylines() ?

_ I think we can replace it. 


if so would it be possible to optimize the existing method instead?

_ It was hard for me to understand why we need interior and exterior and area etc. I couldn't find the usage anywhere else. I think someone who wrote that code can make proper changes to optimize it if he/she see my code algorithm. 

does the new method support disjoint domains? and domains with one or more holes?

_ yes, as you can see in Lake Ontario's Mesh there are several islands. 


could you please explain your algorithm in words (or add a few explaining comments in the code)?

_ the idea is to get all the exterior faces. select the first face as the starting points and look at the remaining faces if the second node of the face is used as the first node of another face, that means that the lines are connected. Once it search through all the faces and no longer meet the condition that means that that polyline is closed. Now the idea is to remove all the faces that was used in this polyline, so next time the code doesn't search through them again. so I have a face_remains variable to handle this. if there are faces remain that means that we still can have another polyline so we repeat this until there is no face remains in face_remains variable.

jsmariegaard:
The primary reason for going through the trouble of finding interior and exterior boundaries is for the contains() method https://github.com/DHI/mikeio/blob/fd32e0b1bf4147d92c23f4883abe9b57b0d66e66/mikeio/dfsu.py#L513 which is used by other methods (and also quite useful itself). 

I have tried the old method on your mesh - and I can see that it does not converge :-( Bad algorithm. 

If you just add the signed area calculation:

    area = area + (xn[n0] - xn[n1]) * (yn[n1] + yn[n0]) / 2 

to your algorithm for each polyline, I think we have all we need :-) If area is positive, then polyline is exterior, otherwise interior. 

Could you do that, @mohm-dhi ?

jsmariegaard:
@mohm-dhi I think this looks great. I do think we should remove the big mesh file again though. I asked for it just for my own testing. I think it is a bit too big for the repository.

mohm-dhi:
I agree to remove the big mesh. Are you using Black for formatting?

jsmariegaard (approved):
Great with an improved algorithm!

jsmariegaard:
@ecomodeller I think this is ready for merge - can you do that?

jsmariegaard:
> I agree to remove the big mesh. Are you using Black for formatting?

Yes. Just installed newest black :-)

mohm-dhi:
> > I agree to remove the big mesh. Are you using Black for formatting?
> 
> Yes. Just installed newest black :-)

I was curious because I am also using Black Version 20.8b1, when I use it the file formatted differently for me. I am using pycharm and have black as a external tools installed to pycharm.

mohm-dhi:
Also could you please fix this before merging 

 e2_to_e3 = np.array(e2_to_e3)

VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  e2_to_e3 = np.array(e2_to_e3)


Thank you

jsmariegaard:
> > > I agree to remove the big mesh. Are you using Black for formatting?
> > 
> > 
> > Yes. Just installed newest black :-)
> 
> I was curious because I am also using Black Version 20.8b1, when I use it the file formatted differently for me. I am using pycharm and have black as a external tools installed to pycharm.

I am using VS Code and have black-formatting-on-save. But I am currently using 3 different machines so version of black is probably different. I have noticed the differences with black before. It is a bit annoying.

jsmariegaard:
> np.array(e2_to_e3)

I was not fast enough - @ecomodeller merged before I committed. But I have fixed it on #156

PR #156, opened 2020-12-21
Dfsu speed direction notebook fix

### LINKED ISSUE #142: Dfsu - Speed and direction  notebook example needs fixing
Include code snippet

**Expected behavior**
Arrows should show. 
There is a simple fix.` dfs.plot()` returns the axis `ax` on which plots. You just need a small change to make it work and use the returned `ax` to call `quiver` not from `plt`:

```
ax = dfs.plot(cs, label="Current speed (m/s)")
ax.quiver(x, y, u, v, scale=6, minshaft=3)
```
The rest of the code can stay the same except delete
```
ax.set_ylim([None, ylim])
ax.set_xlim([xlim, None])
```
which are not only unnecessary but it will prevent proper rendering.

An example figure of quiver that works is attached
LOntario_Vel-crop

It would be nice if we could distribute the velocity vectors on a fictive structured grid to avoid the concentration of vectors in the areas with small elements. Is there any way we can do it now?

jsmariegaard:
Fix #142


jsmariegaard:
@ecomodeller if you are happy with this just go ahead in merge this

PR #157, opened 2020-12-21
Option to hide progress bars

ecomodeller:


mohm-dhi (approved):
It looks very good. I rather prefer to have the default as True.

Thanks.

jsmariegaard:
Would it make sense to:

a) call the option show_progress (instead of hide_progress)? 
b) make the default conditional on the file size: enable for large files? (then most users don't need to worry about show/hide)

mohm-dhi:
Looks great to me. Thanks

> Would it make sense to:
> 
> a) call the option show_progress (instead of hide_progress)?
> b) make the default conditional on the file size: enable for large files? (then most users don't need to worry about show/hide)

a) I like to have show_progress and default as False. 
b) It is not just about file size, sometimes user want to read items in a loop which which end up to have lots of progress bar.

mohm-dhi:
Hi @ecomodeller,
Could you please merge this? Thank you!

PR #164, opened 2021-01-08
Dfs1 and Dfs2 NonEquidistant Time Axis

mohm-dhi:
I noticed that you already change the method find_nearest_element in the dfsu to find_nearest_elements so I applied that change to the dfs2 as well. 

Additionally, as per our discussion I modified the code to handle the NonEquidistant  Time axis as well. When I tried to do that I noticed that some of the checking for Equidistant and nonEquidistant  was some of the codes were there before. in The current code both Equidistant and nonEquidistant Time Axis can be created depend on the parameter that is used by user. 

Thanks,

### DIFF
--- a/mikeio/dfs.py
+++ b/mikeio/dfs.py
@@ -149,6 +149,7 @@ class _Dfs123:
                 raise DataDimensionMismatch()
         if dateTimes:
             self._is_equidistant = False
+            start_time = dateTimes[0]
 
         dfs = self._setup_header(filename)
 
--- a/mikeio/dfs.py
+++ b/mikeio/dfs.py
@@ -122,7 +122,7 @@ class _Dfs123:
 
         dfs.Close()
 
-    def _write(self, filename, data, start_time, dt, items, coordinate, title):
+    def _write(self, filename, data, start_time, dateTimes, dt, items, coordinate, title):
         self._write_handle_common_arguments(
             title, data, items, coordinate, start_time, dt
         )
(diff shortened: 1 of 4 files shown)


jsmariegaard (changes requested):
Great work, @mohm-dhi ! Some minor comments: 

could you please 

* rename dateTimes to datetimes (to make it consistent with Dfs0 class)
* change the order of dt and datetimes (to make it consistent with Dfs0 class)

And a question: 

Could end_time() property be on the base class instead? Is that possible? And could the implementation depend on self._is_equidistant? You don't need to read the file if it's equidistant.

mohm-dhi:
> Could end_time() property be on the base class instead? Is that possible? And could the implementation depend on self._is_equidistant? You don't need to read the file if it's equidistant.

Hi @jsmariegaard 
sure, I will do those changes. 

Regarding you question. The answer for both is Yes, it can be in the base class and in case of equidistant it doesn't need to read the file because it can be calculated from n_timesteps and start_time.

jsmariegaard (approved):

jsmariegaard:
I was a bit fast clicking "accept" :-) 

Could you please let end_time() property be on the base class instead? And please let the code depend on self._is_equidistant? You don't need to read the file if it's equidistant.

Thanks

mohm-dhi:
Hi Jesper,
Sure I will do that but for now, I am very busy with the task I have to finish for the projects.


Thanks,
Mohammad

jsmariegaard:
> Hi Jesper, Sure I will do that but for now, I am very busy with the task I have to finish for the projects. Thanks, Mohammad 

I know that situation all too well :-) No rush.

PR #166, opened 2021-01-10
Dfs3 improvement

mohm-dhi:
Even that dfs3 is not a popular file format that is used in MIKE software but still have some application. I tried to improve the code a little bit by adding the projection property and also have a method that return the bottom values for a 3D dfs3 file.


jsmariegaard (commented):
I never use dfs3 files, but I would think that these files are often huge, so maybe it's not a good idea to read all the data before finding the bottom_data. (but I now see the same is done in read_slice) 

I think it would be best to re-implement read_slice() to handle this situation. read_slice() could have some helper functions that handle the slicing (without reading any data) and then the read() function could be modified to take an argument "elements" or similar for actually reading the data. 

Maybe the existing read_slice() should be renamed read_xy_subset() - imo it does not read a slice. I also think there is a bug in read_slice() as it assumes another sequence of axes as the read() method. I think a read_slice() method should return data corresponding to holding either j-, k- of l-index fixed. 

It may be too much for this PR, though... and as already mentioned: I never use dfs3 files :-)

tests/test_dfs3.py
```diff
@@ -3,38 +3,30 @@
 from datetime import datetime
 import numpy as np
 from mikeio.dfs3 import Dfs3
+from mikeio import Dfsu
```
ecomodeller, line 6: Do you use the Dfsu class in this file?

ecomodeller (approved):

tests/test_dfs3.py
mohm-dhi: No I may just wanted to test something. please remove it.

ecomodeller:
Good to go? Or any more things to change before I merge this?

mohm-dhi:
> Good to go? Or any more things to change before I merge this?

I think it is good to go for this PR. I probably add more features in the future.

Issue #165, opened 2021-01-08
Find Nearest Elements in a Dfsu files


mohm-dhi:
there is a method "find_nearest_elements" that get x,y,z or layer number to return the element index. In case when the z is specified the element index can be wrong when water level is changing dramatically, it this case the search for the element index should be carried out for each time step if requested by the user. In this case the method should return a timeseries of all element indexes for each time step.

jsmariegaard:
Yes, this is true. I have been wondering about how to handle this in a good way. The current functions use only the static information in the file. For many purposes, this will be enough, but for some, it will be wrong. Maybe we should have a separate function called find_nearest_elements_dynamic() or a switch use_dynamic_z in the existing functions? I think the user should actively ask for dynamic_z as it potentially will take a lot longer if you e.g. have a 200GB dfsu file.

mohm-dhi:
To handle this we need to call the "find_nearest_elements" method for each time step and update the z value used in the method based on the changes in the water level which can be calculated from the first item "Z coordinate". 

Below code should work. 

Z_coordinate = dfs.read([0]).data[0]
index = []
initial_index = find_nearest_elements(lon,lat,z)
for t in range(n_timesteps):
    z_new = z + Z_coordinate[0, initial_index] - Z_coordinate[t, initial_index]
    index.append(find_nearest_elements(lon,lat,z_new))

mohm-dhi:
@jsmariegaard 
@ecomodeller 

I was looking into this and found out that the method has an issue when we use layer as an input argument. see this
![image](https://user-images.githubusercontent.com/65772896/104238495-51eee300-5427-11eb-9eef-ffbb6c8a0bc5.png)
for that coordinate the depth is about 85 meter and total number of layers is 31 (10 sigma and 21 z) . it turns out that find_nearest_elements with layer argument just works for the sigma layers and fails when it goes to the z layers.

PR #167, opened 2021-01-11
Fix the issue with find_nearest_elements by layer number

mohm-dhi:
Regarding this comments
https://github.com/DHI/mikeio/issues/165#issuecomment-758227869

I just found what the problem is.


ecomodeller (approved):
Is it possible to add a test to make sure it stays correct?

mohm-dhi:
> Is it possible to add a test to make sure it stays correct?

@ecomodeller I just added a test

Issue #155, opened 2020-12-17
Export mesh/dfs2/dfsu data to shapefiles/geotiffs


georgebv:
**Is your feature request related to a problem? Please describe.**
I want to be able to export mesh/dfs2/dfsu data to GIS formats

**Describe the solution you'd like**
DFSU has `to_shapely` method. I propose adding the following methods (to all dfsu/dfs2 and mesh when its implemented):
- `to_shapefile(path: typing.Union[str, pathlib.Path], item: str, timestep: datetime)` - writes dfsu/dfs2 timestep/item to an ESRI shapefile (polygons of triangular and/or rectangular shapes)
- `to_geotiff(path: typing.Union[str, pathlib.Path], item: str, timestep: datetime, resolution: float)` - creates a raster with resolution in meters with data for given item/timestep

**Describe alternatives you've considered**
Doing it separately, but this would defeat the purpose of using the `mikeio` library. I have already done this via DHI SDK, but having it natively in `mikeio` would be better.

**Additional context**
None.

jsmariegaard:
Hi @georgebv , good points - have you seen the notebook https://github.com/DHI/mikeio/blob/master/notebooks/Dfsu%20-%20Export%20to%20shapefile.ipynb

georgebv:
Hi @jsmariegaard , this looks good. It would be very useful if this was buil-in and accessible via a single method. Often we want to quickly visualize a mesh and having to write boilerplate code is boring.

As I looked at it, I have another suggestion: perhaphs you can add an argument to the proposed `to_shapefile` method called `feature_type`. It would have these options:
- "centroid" - exports points (centers of triangles/quads with elevations as attribute)
- "vertex" - same as "centroid", but exports vertices
- "polygon" - exports polygons (triangles/quads) with elevations as attribute
- "contour" - exports contours with elevations as attribute (for contour you might need additional argument for interval, which would be ignored for other options)

jsmariegaard:
Good ideas. @georgebv would it possible for you to start a PR with some of this stuff?

murray91:
[...]

PR #170, opened 2021-01-30
Exporting Dfsu to Dfs2

georgebv:
This PR introduces a new method to the `Dfsu` class - `.to_dfs2`. This method provides an interface to convert `Dfsu` to `Dfs2` with similar functionality to what DHI MIKE has when creating DFS2 files from DFSU files.

### DIFF
--- a/mikeio/dfsu.py
+++ b/mikeio/dfsu.py
@@ -29,6 +29,7 @@ from .dotnet import (
     asnetarray_v2,
 )
 from .dfs0 import Dfs0
+from .dfs2 import Dfs2
 from .eum import ItemInfo, EUMType, EUMUnit
 from .helpers import safe_length
 from .spatial import Grid2D
@@ -2380,6 +2381,7 @@ class Dfsu(_UnstructuredFile):
         Mesh._geometry_to_mesh(outfilename, geometry)
 
     def to_dfs2(
+        self,
         x0: float,
         y0: float,
         dx: float,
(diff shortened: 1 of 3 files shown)


georgebv:
@ecomodeller @jsmariegaard I cloned the repo and, before making any changes, I used `black` to format the `mikeio/dfsu.py` file, as specified in your contributing guidelines. As you can see, `black` formats the current `dfsu` file - are you OK with this?

I use black all the time on other projects and all the formats made were totally expected. I am just not sure whether this commit won't interfere with other branches you work on.

georgebv:
I was working on #155 and realized that in order to be able to reliably create raster/2D grids the package is currently missing some key components. Namely, we need the following:

1. function/class for conversion between `projection_string` and EPSG (both directions)
2. Way to determine `Dfs2.write` orientation in the `coordinate` argument

To solve 1, I propose making `pyproj` a dependency (it has wheels on PyPI, so installation is not a problem - this should be an easy dependency). Then writing a function to convert between the two should be relatively trivial. I suggest making a class which would define a `mikeio.CRS` object and would have methods (both to- and from-, I show only to-) such as `to_projection_string` (DHI format), `to_epsg` (returns integer code), `to_pyproj`, `to_wkt` (well-known text), etc. Naturally, I cannot do this myself because this requires knowledge of DHI coordinate system nomenclature. @ecomodeller , @jsmariegaard - can you help with this? I can help you create the Class if you tell me how DHI CRS's work.

I found a way to solve 2 when working on DFSU/DFS2 objects for our firm (COWI) internal Python library using MIKE SDK. This solution, however, requires 1 to be solved first (in our library we used a function to convert from UTM string in MIKE to pyproj, non-utm coordinate systems were rejected via ValueError).

jsmariegaard:
Hi @georgebv - have you seen https://github.com/DHI/mikeio/blob/master/notebooks/Dfsu%20-%202D%20interpolation.ipynb which shows how to interpolate a dfsu to an overset grid and to export to dfs2. Doesn't this cover most of it?

georgebv:
@jsmariegaard it doesn't cover grid rotation and CRS, which would be required for raster export - these two features are incompatible with `Grid2D`, which is strictly flat and uses internal DHI notation for coordinate systems. It may be a good idea to update `Grid2D` class to add rotation (orientation) to it. New `CRS` class is required too. Please see my points above

jsmariegaard:
> @ecomodeller @jsmariegaard I cloned the repo and, before making any changes, I used `black` to format the `mikeio/dfsu.py` file, as specified in your contributing guidelines. As you can see, `black` formats the current `dfsu` file - are you OK with this?
> 
> I use black all the time on other projects and all the formats made were totally expected. I am just not sure whether this commit won't interfere with other branches you work on.

Yes, we use black too - but it's probably something with different versions???

georgebv:
I use the latest version, you don't specify version in your contribution guidelines.

jsmariegaard:
> I use the latest version, you don't specify version in your contribution guidelines.

Sure. That's fine. I am not sure if that is the problem at all. I just noticed this in other situations as well.

jsmariegaard:
I've been thinking more on what you wrote on orientation and CRS. I think it's a good idea with some sort of projection class - but I am not the best person to advise on CRS. But as a start, I can provide a few relevant links: 

* https://manuals.mikepoweredbydhi.help/2021/General/MIKE_Zero_Geodesy.pdf 
* https://manuals.mikepoweredbydhi.help/2021/General/Class_Library/DHI_Projections/html/N_DHI_Projections.htm 

The latter is documentation for the Projections assembly which is in the bin folder of mikeio. In here you will also find the Cartography class which is used in the Dfs2 class already https://github.com/DHI/mikeio/blob/ca7f3e6ce73c9e216c99a76c3d813f88a7442e30/mikeio/dfs2.py#L91 

I am not sure, though, why we could not start with to_dfs2() method that just outputs the dfs2 in the same projection as the dfsu and without any rotation and using the overset grid? That would make a good small PR before moving on to the next thing...

georgebv:
> I am not sure, though, why we could not start with to_dfs2() method that just outputs the dfs2 in the same projection as the dfsu and without any rotation and using the overset grid? That would make a good small PR before moving on to the next thing...

Good point. I can raise `NotImplementedError` if epsg or rotation values are passed. Ideally this method should be fully equivalent to the functionality of MIKE itself, when you create dfs2 from dfsu. I think this will require four PR's for that:

1. dfsu to dfs2 skeleton (this PR)
2. CRS class, which translates MIKE projection_strings to common notation (epsg, wkt, etc.)
3. update dfsu to dfs2 to feature full MIKE-compliant workflow (rotation and epsg)
4. dfs2 to raster method

I will try to finish the first one this week. As for number 2, I have some ideas, but it will be a team effort as, in the long run, it may be used by all dfs* classes.

georgebv:
@jsmariegaard I think the PR is ready. I added both method and tests. I also creates `conftest.py` to dynamically dispatch fixtures for it as well. Changes in areas of code outside of functions I created are due to `black` - it appears that either older version of black was used for those sections or they were not formatted.

Let me know if this works. I suggest several next steps, those are listed in my previous comment.

ecomodeller:
Merged

PR #173, opened 2021-02-07
Coordinate Reference System (CRS) class

georgebv:
This PR introduces a new class - `CRS`, which provides the following functionality:

- conversion between DHI MIKE projection string and common coordinate system definitions such as WKT, EPSG, `pyproj` projections, and others
- [will be done in another update] transform spatial data from one coordinate system to another
- [will be done in another update] additional helper methods used to perform spatial calculations by other modules

Motivation to include this class is to provide a way for users to specify custom coordinate systems in which dfs* files are created and to convert existing data to new coordinate systems. Additionally, this class will be used to export dfs* data to vector files (e.g. shapefiles, geopackages) and raster files (e.g. geotiff), because common geospatial libraries such as GDAL and OGR wouldn't recognize native DHI MIKE projection strings. This was originally discussed in https://github.com/DHI/mikeio/pull/170.


georgebv:
Rebased the fork branch against `mikeio` master branch.

georgebv:
@ecomodeller I think the basis of the `CRS` class is ready - we can use it convert to/from `pyproj.CRS`, WKT, and EPSG. Additional methods and aliases can be added in the future as needed.

PR #174, opened 2021-02-08
Refactoring validation of items and time_steps

jsmariegaard:
Refactoring the way we validate requested items and timesteps. no api change. no new functionality.


jsmariegaard:
I have worked a bit to improve the test coverage too (even though it was not really the purpose of this PR)

mikeio/dfs.py
```diff
@@ -58,15 +59,15 @@ def read(self, items=None, time_steps=None):
...
 
-        items, item_numbers, time_steps = self._get_valid_items_and_timesteps(
-            items, time_steps
-        )
+        item_numbers = valid_item_numbers(self._dfs.ItemInfo, items)
+        n_items = len(item_numbers)
+
+        time_steps = valid_timesteps(self._dfs.FileInfo, time_steps)
 
         for t in time_steps:
```
ecomodeller, line 67: Is this still necessary?

mikeio/dfsutil.py
```diff
@@ -0,0 +1,132 @@
...
+    item_lookup = {name: i for i, name in enumerate(names)}
+    try:
+        item_numbers = [item_lookup[x] for x in item_names]
+    except KeyError:
+        raise KeyError(f"Selected item name not found. Valid names are {names}")
+
+    return item_numbers
+
+
+def get_item_info(dfsItemInfo, item_numbers):
```
ecomodeller, line 109: Add default argument to item_numbers (default should be all items)

mikeio/dfs.py
```diff
@@ -13,6 +13,7 @@
 )
 from .eum import ItemInfo, TimeStepUnit, EUMType, EUMUnit
 from .custom_exceptions import DataDimensionMismatch, ItemNumbersError
+from .dfsutil import valid_item_numbers, valid_timesteps, get_item_info
```
ecomodeller, line 16: Rename dfsutil to _dfsutil to indicate it is for internal use?

ecomodeller (approved):

mikeio/dfs.py
jsmariegaard: I assume that you're referring to the lines below:

        for t in time_steps:
            if t > (self.n_timesteps - 1):
                raise IndexError(f"Timestep {t} is > {self.n_timesteps-1}")

which I have now deleted

mikeio/dfsutil.py
jsmariegaard: Done

mikeio/dfs.py
jsmariegaard: Maybe. But perhaps **better to rename the methods to underscore** _variants()? I looked at this https://stackoverflow.com/questions/3602110/python-private-module-in-a-package and https://stackoverflow.com/questions/55157525/leading-underscore-before-the-name-of-python-module  "In the Python standard library, a leading underscore in a module name may denote an accelerator module: a module written in C that provides a more performant implementation of the pure Python version" https://www.python.org/dev/peps/pep-0399/#details

jsmariegaard: agreed to rename methods instead of module (on phone)

PR #185, opened 2021-02-17
new method extract_surface_from_3d

jsmariegaard:
Construct a dfsu 2d file with a surface elevation item, based on zn node coordinates of dfsu 3d file.


jsmariegaard:
test is on the way

PR #194, opened 2021-04-06
Abstract base class as contract for common props

ecomodeller:
To ensure consistent naming of properties across classes.

Now it is clear that:
- `Dataset`
- `Dfs0`
- `Dfs1`
- `Dfs2`
- `Dfsu`

all have the same properties like  `start_time` and `end_time`.


jsmariegaard (approved):
Good idea with these "interfaces"!

PR #197, opened 2021-04-16
missing _source in dfs0

jsmariegaard:


jsmariegaard:
@ecomodeller , do you know why this is suddenly failing in some plot functions with messages such as "This probably means that tk wasn't installed properly"?

ecomodeller:
Add a test to motivate why it is necessary to add a "missing" private variable.

This will probably make it clearer if there is some useful functionality that is missing.

ecomodeller:
FYI `main` is now the default branch

PR #201, opened 2021-04-28
Mikecore

ecomodeller:
Progress so far:
Tests 98 failed 👎 , 180 passed 👍 

Pfs, Mesh, Res1d, xns11 not included 😢

PR #206, opened 2021-06-01
Time index fix

jsmariegaard:
The use of Dfs0Util.ReadDfs0DataDouble() gives problems with rounding errors. 
 
rounding_errors

This can be handled by rounding to milliseconds after converting to DateTimeIndex. This also reduces the need for having to-seconds rounding in to_dataframe() method. 

Furthermore, the previous time conversion with a generator and a list comprehension was found to be slow and has been been replaced by a fast pandas method.

### DIFF
--- a/mikeio/dfs.py
+++ b/mikeio/dfs.py
@@ -97,7 +97,7 @@ class _Dfs123(TimeSeries):
 
             t_seconds[i] = itemdata.Time
 
-        time = [self.start_time + timedelta(seconds=t) for t in t_seconds]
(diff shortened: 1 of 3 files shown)


jsmariegaard:
The changes in this PR reduces the read time of large dfs0 files. A test with a 555MB dfs0 file now loads in 50s instead of 1:35 before. To make it even faster you have to wait for the #201 ...

ecomodeller (approved):
:shipit:

PR #212, opened 2021-06-17
fix extract_track() index bug

jsmariegaard:
extract_track: check for time overlap, fix EOF check


ecomodeller:
If I provide a track where the time is not increasing monotonic, will I get a clear error message with a suggestion on how to fix the input?

jsmariegaard:
> If I provide a track where the time is not increasing monotonic, will I get a clear error message with a suggestion on how to fix the input?

Now, it will say: "AssertionError: The time index must be monotonic increasing. Consider df.sort_index() before passing to extract_track()."

jsmariegaard:
@ecomodeller I think this is ready for merge. What about you? Can I merge it?

PR #214, opened 2021-07-01
Example scripts

ecomodeller:


ecomodeller:
@jsmariegaard If we want to encourage people to use scripts, we better create some examples for inspiration

PR #217, opened 2021-08-17
Equidistant vs non-equidistant

ecomodeller:


jsmariegaard (changes requested):
Looks good. Except for old text in notebook: "The write method allows to modify the data without the need of specifying names, units etc. If you need to add variables, rename, change units, you must use create instead." - and maybe a heading in the notebook.

mikeio/dataset.py
```diff
@@ -148,6 +148,8 @@ def __repr__(self):
         out = ["<mikeio.DataSet>"]
         out.append(f"Dimensions: {self.shape}")
         out.append(f"Time: {self.time[0]} - {self.time[-1]}")
+        if not self.is_equidistant:
+            out.append("-- Non-equidistant time axis --")
```
jsmariegaard, line 152: non equidistant *calendar* axis?

PR #223, opened 2021-08-25
Generic Time average

ecomodeller:
As discussed here: https://github.com/DHI/mikeio/discussions/196


ecomodeller:
It would be nice to accept any aggregation function, but it is difficult to do while avoiding reading the entire dataset in memory.

One thing to consider, should this be in the `generic` or the `aggregator` module?

PR #226, opened 2021-08-25
Dataset math operations

jsmariegaard:
Allow basic math operations +, - and * on Datasets. Both with scalar and other Datasets.


ecomodeller:
Dont't forget https://docs.python.org/3/reference/datamodel.html#object.__radd__
https://stackoverflow.com/questions/42071861/python-commutative-operator-override

```
ds + 1.0 == 1.0 + ds
```

PR #227, opened 2021-08-25
Dfs0 end_time property. Closes #203

### LINKED ISSUE #203: end_time property does not work in dfs0
https://github.com/DHI/mikeio/blob/ed60b7caeb1e2a2003266a69607157f2f3302c20/mikeio/dfs0.py#L424

Will throw an error since _end_time is not defined. Its implementation should be similar to that in dfs.py as it depends on the type of TimeAxis.

ecomodeller:


jsmariegaard:
My suggestion:

    if self._source.FileInfo.TimeAxis.IsEquidistant():
        dt = self._source.FileInfo.TimeAxis.TimeStep
        n_steps = self._source.FileInfo.TimeAxis.NumberOfTimeSteps
        timespan = dt*(n_steps-1)
    else:
        timespan = self._source.FileInfo.TimeAxis.TimeSpan
    
    end_time = self.start_time + timedelta(seconds=timespan)

PR #244, opened 2021-09-21
Missing EUM types and units

ecomodeller:


ecomodeller:
nanomole_per_cubic_ounce_per_lightyear ...

Supporting:
* SI
* Imperial
* Prefix
* Combined units

No wonder it gets out of sync 😓

PR #262, opened 2021-11-10
Does it run on 3.10?

ecomodeller:


ecomodeller:
Since there are no pyproj wheels available on pypi for Python 3.10 at the moment, MIKE IO can not be installed on 3.10 if pyproj is listed as a dependancy.

We will treat pyproj as optional, similar to shapely and if you need any of these, you can install them from 
[Christoph Gohlke site](https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyproj)

PR #264, opened 2021-11-12
Add Dfs2 x0 y0 properties

jsmariegaard:
x0,y0 was up to now read from Dfs2 files. In most cases (2d horizontal) these will be 0, but for point spectra files at least the frequency axis will not start at 0. 

I know this interests @Hendrik1987


ecomodeller (approved):
Is x0, y0 only relevant for Dfs2? 

What about Dfs1 and Dfs3?

ecomodeller (approved):
Is x0, y0 only relevant for Dfs2? 

What about Dfs1 and Dfs3?

Hendrik1987:
Thanks @jsmariegaard  for picking this up, exactly what I was missing :-)
@ecomodeller, I think you are right and x0, y0 might be relevant for vertical current profiles in Dfs1 for example - but I have not tested this in mikeio yet.

PR #272, opened 2022-01-04
DataArray

ecomodeller:
This PR introduces a new abstraction, the DataArray.

A DataArray has the data in the form of a numpy array, but also a name and and EUM type, and a geometry which will make plotting the data along with sensible labels very easy.

I.e. a Dataset consists of one or more DataArrays.

**Wanna try it out?** 

Option 1:

    > pip install --upgrade https://github.com/DHI/mikeio/archive/refs/heads/dataarray.zip

Option 2: 

Run [this notebook](https://colab.research.google.com/drive/1Nevw-nhZKwHR4FYZmGSWoZuLfyaDQYn4) in Colab (requires google account)


ecomodeller:
DataArray indexing returns a new DataArray 🚀

ecomodeller:
How about returning a DataArray if the items argument is scalar?

That would avoid writing the item name twice.
image

But indexing like this is rather short
image

🤔

jsmariegaard:
Yes, we should allow both - you use the same in FMskill - it's quite easy to implement

jsmariegaard:
> Yes, we should allow both - you use the same in FMskill - it's quite easy to implement

Something like:

    def __getitem__(self, x):
        if isinstance(x, int):
            x = self.items[x]

jsmariegaard:
Slicing would also be nice ds[0:2] ...

ecomodeller:
> But why?

Static analysis is useful when writing scripts:
If `mikeio.read()` always returns a `Dataset`, we get useful command completion!
image

jsmariegaard:
> > But why?
> 
> Static analysis is useful when writing scripts: If `mikeio.read()` always returns a `Dataset`, we get useful command completion! image

Could point. But how about issuing helpful error messages for xyz and mesh files...

mikeio/dfs0.py
```diff
@@ -1,6 +1,7 @@
 import os
 import warnings
 from datetime import datetime, timedelta
+from h11 import Data
```
jsmariegaard: WTF?

PR #276, opened 2022-01-14
Generic concatenation, keep first or last

clemenscremer:
Added keep argument and functionality to function concat.
Keep now allows specification which data to keep "first" (older) or "last" (newer), as demonstrated in attached image.

image


ecomodeller (changes requested):
It looks promising, but it needs verification with tests.

The existing test is not complete, since it doesn't verify the behavior in the overlapping period.
https://github.com/DHI/mikeio/blob/2acbdb48575b6ca477ea59ba7205eb1174fc97b9/tests/test_generic.py#L231

or add a new one demonstrating the effect of the new keep option.

I.e. assert that at least some of the data in an overlapping period comes from the first or the last. The existing test with data from the tidepredictor, is probably not very useful in this case, since they are identical in the overlapping period, so you have to create some new small test data.

ecomodeller (approved):

PR #290, opened 2022-03-11
Grid 2D create with nx, ny

ecomodeller:
Creating a Grid2D like this:
```
Grid2D(shape=(10,2),..)
```
is intuitive with `shape=(nx,ny)`, except that it matches poorly with the shape of the data where the `shape=(nt,ny,nx)` 😕 

Thus in the spirit of explicit is better than explicit (and a Grid2D doesn't accept anything else than 2d anyway):
```
Grid2D(nx=10, ny=2,...)
```


jsmariegaard:
Agree let's remove the shape argument

PR #292, opened 2022-03-16
Datarray/Dataset interp_like

ecomodeller:
TODO 
* accept DataArray and Dataset as arguments
* interp_time

### DIFF
--- a/mikeio/dataarray.py
+++ b/mikeio/dataarray.py
@@ -1040,27 +1040,32 @@ class DataArray(DataUtilsMixin, TimeSeries):
         """
 
         if hasattr(other, "geometry"):
-            grid = other.geometry
+            geom = other.geometry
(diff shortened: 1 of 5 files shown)


jsmariegaard:
Should it also be allowed to just _interpolate in time_ with this method? 

    ds.interp_like(ds2.time)

and maybe ds and ds2 have the _same_ geometry and the user actually only wants to interpolate in _time_ using: 

    ds.interp_like(ds2)   # where ds.geometry = ds2.geometry

jsmariegaard:
Now added DatetimeIndex to possible "other" in interp_like - let me know what you think @ecomodeller

jsmariegaard (approved):
See comments and changes - let's postpone adding bilinear interp for grid2d to another PR

PR #294, opened 2022-03-22
Fix setup.py

daniel-caichac-DHI:
there was an additional `.` at the end of Numpy (`1.15.0.`) that created problems in conda, not allowing to create/export environments, and maybe some other problems.


ecomodeller:
That is a strong contestant for the worlds smallest PR 🔬. But we'll take it in anyway 😉

PR #295, opened 2022-03-22
Fix spectral dfsu

jsmariegaard:
It should be possible to select specific elements/nodes from spectral dfsu files


ecomodeller (approved):
I'm still missing frequency and direction values available on the Dataset/DataArray.

jsmariegaard:
> I'm still missing frequency and direction values available on the Dataset/DataArray.

Agree - let's do it in a later PR - we also still miss more natural plotting of spectra e.g. da_spectra.plot() - which could be re-used for dfs2 spectra.

PR #300, opened 2022-03-25
Dfsu3d extract column

jsmariegaard:
Better extraction and plotting of vertical columns. 

image


jsmariegaard:
Stuff to consider: 

* The read() method has for long had the argument `elements`. This PR adds a new argument `layers` - the plural form is consistent with element**s** but inconsistent with the `layer` argument in Dataset/DataArray. What to do? use `layers` or `layer`?

jsmariegaard:
@ecomodeller, the failing test (which I have now disabled) is a pandas indexing issue that I think we should fix in another PR.

ecomodeller:
> @ecomodeller, the failing test (which I have now disabled) is a pandas indexing issue that I think we should fix in another PR.

I have added a new failing test that is related 😉 (Data is no longer 2d after choosing a specific timestep)

ecomodeller (approved):

PR #303, opened 2022-03-29
Default to float32, and allow float64

ecomodeller:
Data is typically stored as float32 (single precision). Change default behaviour to read without conversion to float64 unless explicitly requested.

### DIFF
--- a/tests/test_dfsu.py
+++ b/tests/test_dfsu.py
@@ -70,13 +70,16 @@ def test_read_single_precision():
     assert ds[0].dtype == np.float32
 
 
-def test_read_double_precision_open():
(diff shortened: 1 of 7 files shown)


tests/test_dfsu.py
```diff
@@ -70,6 +70,16 @@ def test_read_single_precision():
     assert ds[0].dtype == np.float32
 
 
+def test_read_double_precision_open():
+    filename = "tests/testdata/HD2D.dfsu"
+    dfs = mikeio.open(filename, dtype=np.float64)
```
jsmariegaard, line 75: maybe add a line where the default dtype is used

PR #307, opened 2022-03-30
add geometry when combine datasets

mohm-dhi:
add geometry to the combined dataset, otherwise _to_dfs() method with throw error, as it looks for geometry.


jsmariegaard:
Thanks @mohm-dhi - this was on my today list. Let's make a quick fix now and I will then revisit the combine, concat workflow next week again :)

jsmariegaard (approved):
Looks good - larger review of combine/concat workflow needed later

PR #308, opened 2022-04-06
Grid{1D,2D} consistency

ecomodeller:
Grid1D should accept a list of coordinates

### DIFF
--- a/mikeio/spatial/grid_geometry.py
+++ b/mikeio/spatial/grid_geometry.py
@@ -9,7 +9,8 @@ from ..eum import EUMType, EUMUnit
 class Grid1D(_Geometry):
     def __init__(
         self,
-        x0=None,
+        x=None,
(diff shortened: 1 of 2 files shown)


mikeio/spatial/grid_geometry.py
```diff
@@ -22,15 +23,17 @@ def __init__(
...
-            raise ValueError("n must be provided")
-        if dx is None:
-            raise ValueError("dx must be provided")
-        self._nx = n
-        self._dx = dx
-        self._x0 = 0 if x0 is None else x0
-        x1 = self._x0 + dx * (n - 1)
-        self._x = np.linspace(self._x0, x1, n)
+        if x is not None:
+            self._x = np.array(x)
```
jsmariegaard, line 27: maybe need to set self._nx here as well?

PR #309, opened 2022-04-08
Spectral geometry classes

jsmariegaard:
Improve the workflow for spectral dfsu files.

- [x] Spectral geometries 
- [x] Dedicated DataArray plotters for context-aware plotting of spectral data
- [x] Fix sel() for spectral data
- [x] add frequencies and directions as attributes on Dataset and DataArray

image

image

Known issues: 

* ax argument cannot be used in full spectrum plot
* Hm0 calculation is not correct for Action Density. Make it so and/or check on EumType before calc.
* Hm0 calculation is not checked for radian/degrees units - should add test and scale


ecomodeller (commented):
X and Y labels, colorbar is missing in the plot:
image

Compare to the xarray version:
image

Not supported methods should return a understandable error message:
image

jsmariegaard:
@ecomodeller, reported problems have now been fixed. Should we merge?

PR #310, opened 2022-04-09
Plot vertical transect

jsmariegaard:
Make it easier to plot vertical transects

image

Known issues with vertical transects (which is out of scope for this PR): 

* selecting multiple elements to new GeometryFMVerticalProfile seems to fail
* sel() area argument does not work 
* sel() layer argument does not work 
* sel() z, y, z argument does not work 
* geometry.plot() not implemented for GeometryFMVerticalProfile


ecomodeller (commented):
1. Would be to be able to add annotations at specific locations
2. And to easy create an horizontal view of where the transect was extracted (compare inset map) 

image

Like this:
image

PR #311, opened 2022-04-09
Grid1D and Grid2d subsetting

jsmariegaard:
* [x] Grid1D optional node_coordinates
* [x] Grid1D.isel() w multiple (equidistant) index return new Grid1D
* [x] Grid2D.isel() w multiple (equidistant) index return new Grid2D
* [x] DataArray with Grid2D: sel()/isel() methods will return PointGeometry with correct coordinate info (thanks to _nc on Grid1D)
* [x] Dfs2 read method to take area argument 
* [ ] Dfs2 read method to take j- an k- index arguments

Things to consider:

* Should we generally use the names j- and k- to refer to the first and second index following MIKE standard? 
* Should the Grid1D node coordinates be public
* polygon in area selection should crop to its bounding box and then making everything outside the polygon np.nan. In that way it would be simple to calculate statistics only over water. 
* What to do with origo and axis in 2dv dfs2 files? They are still common in MIKE3 modelling. TODO: add one as test data
* Remove origo from x0,y0 ? 
* Special class Grid2DRotated() 

image

image

### DIFF
--- a/mikeio/dataarray.py
+++ b/mikeio/dataarray.py
@@ -795,13 +795,15 @@ class DataArray(DataUtilsMixin, TimeSeries):
     ) -> "DataArray":
 
         # TODO: delegate select in space to geometry
+        t_ax = 1 if self.dims[0][0] == "t" else 0
 
         # select in space
         if (x is not None) or (y is not None) or (z is not None):
             if isinstance(self.geometry, Grid2D):  # TODO find out better way
                 i, j = self.geometry.find_index((x, y))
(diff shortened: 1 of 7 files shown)


ecomodeller (approved):
Super!

It would be even better if the `.sel` could accept an area argument, to be consistent with dfsu files 
image

jsmariegaard:
> Super!
> 
> It would be even better if the `.sel` could accept an area argument, to be consistent with dfsu files image

Oops. It can actually, but only DataArray (inconsistent with Dataset). Fixing now.

PR #316, opened 2022-04-26
Fix dropna method

miab44:
Enabled the dropna method to drop nan values inside multiple items.


jsmariegaard:
Thanks @miab44 for fixing this! I have added dropna for DataArray and added tests.

PR #324, opened 2022-05-01
Fix contourf wrong color with discrete non-equidistant colorscale

ecomodeller:


ecomodeller:
Comparison of default plot (patch) and contourf with a discrete colorscale.

Before:
before

After:
after

jsmariegaard (commented):
It may fix the problem, but it also makes vmin and vmax arguments non-functional! They need to work as in patch-plot.

jsmariegaard:
@ecomodeller could this work?

PR #333, opened 2022-05-05
Fix dfs2 read area

ecomodeller:


jsmariegaard (changes requested):
Great stuff. Could you please make Grid3D a dataclass as well - for consistency. And remove its x0,y0,z0 properties. 

Grid1D. maybe remove its x0, x1 properties too? For consistency... 

You added a notebook Untitled.ipynb - is that on purpose? 

And regarding Grid2D: maybe remove _x_logarithmic, _is_spectral, _shift_origin_on_write from dataclass annotations in the beginning (but keep them in the __init__ ) as they are derived from the other and should not be accounted  for in __eq__ etc.

jsmariegaard (approved):

PR #341, opened 2022-05-10
Consistent time parsing

### LINKED ISSUE #340: mikeio.read string time argument not working as intended
mikeio.read("ts.dfs0", time="2018") gives different results for equidistant and non-equidistant files. 

Reading non-equidistant files works as intended (i.e. gives all data in 2018 in the above example). If ts.dfs0 is equidistant, however, only one timestep i returned: 2018-1-1 00:00:00 !

In other cases, an error is thrown when time is a string! The below lines should give the same Dataset, but the second throws an error: 

    mikeio.read("../tests/testdata/HD2D.dfsu")["1985-08-06"]
    mikeio.read("../tests/testdata/HD2D.dfsu", time="1985-08-06")

ecomodeller:


jsmariegaard (changes requested):
Looks good, but I managed to add a few failing tests: 

1. Check on shape reveals that time dimension is not removed consistently. See test_read_dfs_time_selection_str_specific 
2. time as pd.Timestamp is only partly supported
3. time as pd.DatetimeIndex is only partly supported
4. The old comma separated datetime slice "2018-03-08,2018-03-09" is supported by the read() method but not sel() (it is also supported in generic.extract - see below). This should be consistent. The notation is nice for backward compability but inconsistent with xarray.

And what about the start_time argument in the generic.extract() method? It also supports the comma separated string https://github.com/DHI/mikeio/blob/e7a3b0b50cea303e8e02190b6e329e5f1d4ab98d/mikeio/generic.py#L518 - should we remove this?

ecomodeller:
New argument in `mikeio.read(...keepdims=True)` which keeps singleton time dimension

jsmariegaard:
I think the following should be allowed as time input t in read(time=t), da[t] and ds.sel(time=t): 

For selecting single timesteps:

* int
* str
* datetime
* pd.Timestamp

For selecting multiple timesteps: 

* slice of int, str, datetime or pd.Timestamp
* sequence of int, str, datetime or pd.Timestamp
* pd.DatetimeIndex

jsmariegaard:
> :shipit: ?

:shipit:

PR #346, opened 2022-05-18
Error when creating a dataset with many items

ecomodeller:


ecomodeller:
The current validation of compatible item can not handle many items due to recursive code (stack gets too deep).

The easy "fix" is to avoid validation per default, Dfs files written by a MIKE engine should not need validation anyway.

Validation can be turned on like this: `mikeio.Dataset([da1, da2], validate=True)` but will fail if you try to create many items.

jsmariegaard (changes requested):
Looks really good! But please skip also the validation on Dataset transformation methods such as ds.squeeze() etc.

PR #359, opened 2022-06-07
Fix poor performance of read of many-item-dfs0

### LINKED ISSUE #358: Poor performance (or stuck?) in Dfs0 with many columns
**Describe the bug**
Hi Everyone. So this is the problem that I've encountered.
I have a dfs0 with just a few timesteps (10 timesteps of 1hour, so 10 rous), however a bunch of columns (~6500 columns)
When I open the file with MikeIO v`0.12.2` it takes less than a second (I am reading and converting to dataframe, but for this bug report I am just reading it).
Then on the same machine, with the same environment and everything (same pandas, same numpy, etc.) I am upgrading to MikeIO v`1.0.0` and rerunning the notebook, and it takes a **Long time** to open the file. I have been waiting +10 minutes and it is still computing, so maybe it is stuck in a loop or some infinite recurrence, or maybe it is just very slow. 
I can provide the dfs0 as it only weighs ~1mb.

**To Reproduce**
`from mikeio import Dfs0 `
`import mikeio`

`file= r'/Dummy_points.dfs0'`
`ds=Dfs0(file).read()`

**Screenshots**
image

image

**System information:**
 - Python version `3.8.12`
 - Mike IO version `0.12.2` (works) and `1.0.0` (does not work)

Thanks

jsmariegaard:
Closes #358 

The overhead of creating DataArrays for every item seems to be too big for Dfs0s with many items (e.g. +1000 items).

### DIFF
--- a/mikeio/dataset.py
+++ b/mikeio/dataset.py
@@ -311,10 +311,28 @@ class Dataset(DataUtilsMixin, TimeSeries, collections.abc.MutableMapping):
     @staticmethod
     def _check_all_different_ids(das):
         """Are all the DataArrays different objects or are some referring to the same"""
-        for j, da1 in enumerate(das):
-            for k, da2 in enumerate(das):
-                if j != k:
-                    Dataset._id_of_DataArrays_equal(da1, da2)
+        ids = np.zeros(len(das), dtype=np.int64)
+        ids_val = np.zeros(len(das), dtype=np.int64)
+        for j, da in enumerate(das):
+            ids[j] = id(da)
+            ids_val[j] = id(da.values)
+
+        if len(ids) != len(np.unique(ids)):
+            # DataArrays not unique! - find first duplicate and report error
+            das = list(das)
+            u, c = np.unique(ids, return_counts=True)
+            dups = u[c > 1]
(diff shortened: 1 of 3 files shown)


jsmariegaard:
Maybe, we should a to_pandas method to the mikeio root? 

    > df = mikeio.to_pandas("my.dfs0") 

Would make a cleaner syntax than the current: 

    > df = mikeio.Dfs0("my.dfs0").to_pandas()

daniel-caichac-DHI:
> 

This solves the issue as far as I'm concerned. 
Thanks

jsmariegaard:
I've identified two bottlenecks for creating many-DataArray-Datasets: 

* _set_name_attr() which calls _to_safe_name() 
* _check_all_different_ids which uses a double enumerate-loop 

Both can be fixed. This would solve the problem more generally, not forcing users to avoid the mikeio.Dataset. I'll check in a solution later today.

jsmariegaard:
Reading to Dataset with many items (up to 10_000) is now also fast :-)

ecomodeller:
image

ecomodeller (approved):

PR #365, opened 2022-06-10
Interp_like with time invariant data arrays

ecomodeller:


jsmariegaard (changes requested):
Looks good! Found only one small problem: DataArray.interp_like and Dataset.interp_like crash in different ways if I try to interpolate from Grid2D to something else. The first throws a NotImplementedError (L1590), the second complaints about the non-existence of self.geometry.get_2d_interpolant (L1130). I would suggest to include 

        if not (isinstance(self.geometry, GeometryFM) and self.geometry.is_2d):
            raise NotImplementedError("Currently only supports 2d flexible mesh data!")

Also in Dataset.interp_like()

PR #369, opened 2022-06-20
Add new method Dataset.from_xarray

ecomodeller:
Conveniently convert 2d gridded datasets from xarray to mikeio.

Note that this method assumes that all variables have the same dimensions, so first select relevant items before converting. This is easy to do with xarray, hence the method is from_xarray and not from_netcdf.


ecomodeller:
Missing:
- [ ] Option to flip y-axis, many NetCDF files have a descending y-axis, which is not possible in Dfs2.

PR #371, opened 2022-06-23
Fix reading vertical column in Dfsu3d

ecomodeller:


clemenscremer (approved):
all works fine for basin_3d.dfsu and more complex sigma layered dfsu with triangular and quadrangular elements

PR #373, opened 2022-06-23
Save vertical profile from sigma only dfsu 3d

ecomodeller:
Right after releasing v 1.0.2 I found another problem with profiles extracted from DfsuSigma files where number of sigma layers where incorrect and prevented saving them as a new Dfsu file.
image

### DIFF
--- a/mikeio/spatial/FM_geometry.py
+++ b/mikeio/spatial/FM_geometry.py
@@ -1258,11 +1258,8 @@ class GeometryFM(_Geometry):
             else:
                 geom._type = self._type
                 geom._n_layers = n_layers
-                if self._type == DfsuFileType.Dfsu3DSigma:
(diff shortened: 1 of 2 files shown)


mikeio/spatial/FM_geometry.py
```diff
@@ -1258,8 +1258,11 @@ def elements_to_geometry(
             else:
                 geom._type = self._type
                 geom._n_layers = n_layers
-                lowest_sigma = self.n_layers - self.n_sigma_layers + 1
```
jsmariegaard, line 1261: wouldn't be enough to change this line to (remove the +1):

    lowest_sigma = self.n_layers - self.n_sigma_layers

jsmariegaard:
:shipit:

PR #380, opened 2022-06-28
Dataset aggregation across items

jsmariegaard:
Extend existing Dataset aggregation methods (mean, max, aggregate, quantile, etc) to allow for `axis="items"` argument which aggregates across items. A DataArray will be returned by default, but the user can insist on a Dataset by specifying `keepdims=true`. 

Also, add missing dataarray methods: `average` and `nanquantile` (for consistency with dataset). And add "missing" aggregation methods `std` and `nanstd` on both dataarray and dataset. Furthermore, better support for kwargs passing and optional new name for the aggregated item.


jsmariegaard:
@ecomodeller notice that when pressing "ready for review" the full test suite is NOT run. Wasn't that the intention?

ecomodeller (approved):

PR #390, opened 2022-07-13
add time property to mikeio.open and add test

mohm-dhi:
It was often that we only need all the datetimes in the file without reading the entire file or any item. I just add that property

### DIFF
--- a/mikeio/dfs.py
+++ b/mikeio/dfs.py
@@ -1,5 +1,6 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 from abc import abstractmethod
+from typing import Optional
 
 import warnings
 import numpy as np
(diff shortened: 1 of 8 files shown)


jsmariegaard:
Great idea - I have thinking about the same. We just need it on dfs0, 1, 2, and 3 also for consistency. At least in dfs0, we should handle non-equidistant time axis also.

mohm-dhi:
@jsmariegaard I am sorry, I was very busy that could not work on this. I hope to find some time next week to make the necessary changes.

jsmariegaard:
Great. No rush.

mohm-dhi:
> pd.DatetimeIndex

output datetimes are in pd.DatetimeIndex, I added an assert to check for that too.

mohm-dhi:
> Great idea - I have thinking about the same. We just need it on dfs0, 1, 2, and 3 also for consistency. At least in dfs0, we should handle non-equidistant time axis also.

I just added the time property to dfs0 equidistant and non-equidistant time axis, will do the dfs 1, 2 and 3 later

PR #399, opened 2022-08-05
DataArray/Dataset ptp (range)

ecomodeller:
Range (max - min) is an intuitive measure of dispersion, but since range already has a different meaning in python and NumPy, Numpy uses the name [ptp](https://numpy.org/doc/stable/reference/generated/numpy.ptp.html) instead which is now available as a method on the DataArray and Dataset classes.

Below is a illustration of range vs standard deviation (another popular measure of dispersion)
image


jsmariegaard:
❤️ :shipit:

PR #408, opened 2022-08-27
DataArray apply and Numpy array container

jsmariegaard:
New method `apply` similar to methods in pandas and xarray for applying a mathematical function to all values in DataArray. 

        >>> da.apply(np.sin)
        >>> da.apply(np.clip, 0.0, 1.0)
        >>> da.apply(lambda x: x**2 if x > 0 else 0)

### DIFF
--- a/docs/dataarray.md
+++ b/docs/dataarray.md
@@ -110,16 +110,27 @@ including different ways of *selecting* data:
 * [`aggregate()`](DataArray.aggregate) - Aggregate along an axis
 * [`quantile()`](DataArray.quantile) - Quantiles along an axis
 
-*Mathematical operations* +, - and * with numerical values:
+*Mathematical operations* +, -, *, **, / with numerical values:
 
 * ds + value
 * ds - value
 * ds * value
+* ds ** value
+* ds / value
+* abs(ds)
+* -ds
 
-and + and - between two DataArrays (if number of items and shapes conform):
+And apply with a function to all values:
+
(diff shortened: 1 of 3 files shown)


ecomodeller:
Would be even nicer if we could support this syntax;
```python

da2 = np.sqrt(da)
```
I can't really figure out how this works, but somehow it calls the `__array_ufunc__` function.
https://github.com/pydata/xarray/blob/3d3b236df3f8f5982760e68e064eb0db2aacc4a2/xarray/core/arithmetic.py#L43

jsmariegaard:
> Would be even nicer if we could support this syntax;
> 
> ```python
> da2 = np.sqrt(da)
> ```
> 
> I can't really figure out how this works, but somehow it calls the `__array_ufunc__` function. https://github.com/pydata/xarray/blob/3d3b236df3f8f5982760e68e064eb0db2aacc4a2/xarray/core/arithmetic.py#L43

Great idea! A hacky first attempt is now checked in. 

Note also this medium write up about someone attempting something similar: https://towardsdatascience.com/wrapping-numpys-arrays-971e015e14bb

ecomodeller:
This seem like the proper reference
https://numpy.org/devdocs/user/basics.dispatch.html#writing-custom-array-containers

jsmariegaard:
This PR will have to wait some weeks until we have time to do a proper implementation of Numpy array container such that a mikeio.DataArray can be considered truly array-like

ecomodeller:
Interesting, but not for now.

PR #409, opened 2022-08-29
DataArray argmin/argmax

jsmariegaard:
To find the index of min/max values


ecomodeller:
More to add here or :shipit: ?

jsmariegaard:
> More to add here or :shipit: ?

I don't have more to add here. What do you think?

ecomodeller:
> > More to add here or :shipit: ?
> 
> I don't have more to add here. What do you think?

That I don't like failing tests...

The quick test running on Python 3.9 works fine, but the test on 3.7 fails 😢 

The last version of NumPy for 3.7 is 1.21.6 released Apr 12, 2022

jsmariegaard:
Apparently, not ready after all as it does not work on py3.7. Further development on hold for a few weeks.

jsmariegaard:
Will close this as the need is not obvious.

Issue #413, opened 2022-09-05
Performance of interp_like when converting dfsu to dfs2 for large files 


ryan-kipawa:
**Is your feature request related to a problem? Please describe.**
Converting large dfsu files to dfs2 seems to take a long time (for a 120mb item, it took roughly 6 hours). I used the following code:

```
ds = mikeio.read(dfsu_filename, items=item, time=time)
da = ds[item]
g = da.geometry.get_overset_grid(dx=dxdy)
da_grid = da.interp_like(g)
```

The PC has 64GB memory and it's CPU is: Intel Xeon E5-2650 v4 @ 2.20Ghz.

**Describe the solution you'd like**
Is there anything I could do to make this run faster?

**Describe alternatives you've considered**
I haven't tried anything yet. One idea is to vectorize the for loop in "_interp_itemstep".

**Additional context**

jsmariegaard:
You could also try with 

    da_grid = da.interp_like(g, n_nearest=1)

which use nearest point instead of IDW interpolation.

jsmariegaard:
Hmm, yes, the contains() method (to find out if points are inside the mesh) is actually what takes the most time in this case. But closely followed by the dot product. It may be better to set extrapolate=True and then remove points outside the domain afterwards

    > da_grid = da.interp_like(g, extrapolation=True)   # will not call contains()

cethceth:
I have a similar size dfsu file and have the same issue, but I get an error when I try to set 'extrapolation=True':
![image](https://user-images.githubusercontent.com/117161494/199225072-7954c8db-bb52-418b-ad90-deb28a551cb3.png)
Any idea how to solve this?

cethceth:
.. and can it really be true that this process takes that long in Python when in Mike Zero it is quite fast?

jsmariegaard:
@cethceth, which version of MIKE IO are you using?

cethceth:
Solved it with 'extrapolate' instead of 'extrapolation' :)

ecomodeller:
> .. and can it really be true that this process takes that long in Python when in Mike Zero it is quite fast?

Are you referring to Create new Grid series - From Dfsu file (with the less than intuitive grid specification)?
![image](https://user-images.githubusercontent.com/614215/199556594-568e354d-5e58-488e-af6e-8dee41a2e137.png)

cethceth:
> 

[...]

PR #414, opened 2022-09-05
Interpolation matrix multiply (testing...)

ecomodeller:
Vector multiplication in loop replaced with matrix multiplication. (See #413)

Still left to do: create weights matrix without loop.

### DIFF
--- a/mikeio/interpolation.py
+++ b/mikeio/interpolation.py
@@ -120,16 +120,16 @@ def _interp_itemstep(data, elem_ids, weights=None):
         # nearest neighbor
         return data[elem_ids]
     ni = len(elem_ids)
-    #idat = np.empty(ni)
-    #for j in range(ni):
+    # idat = np.empty(ni)
+    # for j in range(ni):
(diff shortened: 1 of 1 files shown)


jsmariegaard:
I think it's better just to replace the whole thing with:

    def _interp_itemstep(data, elem_ids, weights=None):
        if weights is None:
            return data[elem_ids]
        else:
            return np.sum(data[elem_ids]*weights, axis=1)

This is equivalent to taking the dot product of each row but much much faster.

jsmariegaard:
Closer investigation shows that this is needed: 

    def _interp_itemstep(data, elem_ids, weights=None):
      if weights is None:
        return data[elem_ids]
      else:
        idat = data[elem_ids] * weights
        return np.sum(idat, axis=1) if weights.ndim == 2 else idat

To also handle the cases when n_nearest=1...

jsmariegaard:
You closed this? But wouldn't it be good to implement my suggestion?

ecomodeller:
Sure, I'll create a new branch with a fresh start

PR #415, opened 2022-09-06
Faster interpolation

ecomodeller:


ecomodeller:
Interpolation to a structured grid is now faster, so now the `.contains()` is the bad guy.

Default:
`da.interp_like(grid)`

Faster with extrapolation everywhere:
`da.interp_like(grid, extrapolate=True)`

Faster with limited extrapolation:
`da.interp_like(grid, extrapolate=True, radius=0.1)`

jsmariegaard:
Will close #413

PR #417, opened 2022-09-06
Allow asterisks in item names

ecomodeller:
Please comment if you think this a good or bad idea.


jsmariegaard:
What about this? 

    if key in self._data_vars.keys():
        return self._data_vars[key]
    elif "*" in key:
        import fnmatch

        data_vars = {
            k: da
            for k, da in self._data_vars.items()
            if fnmatch.fnmatch(k, key)
        }
        return Dataset(data=data_vars, validate=False)
    else:
        raise IndexError("...")

PR #418, opened 2022-09-08
Remove deprecated functionality

ecomodeller:


jsmariegaard (changes requested):
As discussed: Generic.ipynb needs updating. 

When I search for "FutureWarning" I get this additional deprecated functionality: 

* Dfs0.to_dfs0(): "Dataset.to_dfs0() is deprecated, use Dataset.to_dfs()"
* Dfs3.write(): start_time, datetimes, items deprecated; setting data as list of numpy arrays...  
* Dfsu: codes, find_nearest_elements()
* Dfsu.write(): start_time, dt, items deprecated; setting data as list of numpy arrays...  
* Dfsu.plot()
* DfsuLayered.read(): time_steps argument
* DfsuLayered.get_layer_elements(): layer argument

What are your thoughts on these? Have you not seen these or have you decided we should postpone their retirement? 

I think dfsu and dfs3 write methods should be consistent with the other write methods (and therefore should have these removed). It would also be good to get rid of the other deprecated functionality listed above IMO.

ecomodeller:
> As discussed: Generic.ipynb needs updating.
> 
> When I search for "FutureWarning" I get this additional deprecated functionality:
> 
> * Dfs0.to_dfs0(): "Dataset.to_dfs0() is deprecated, use Dataset.to_dfs()"
> * Dfs3.write(): start_time, datetimes, items deprecated; setting data as list of numpy arrays...
> * Dfsu: codes, find_nearest_elements()
> * Dfsu.write(): start_time, dt, items deprecated; setting data as list of numpy arrays...
> * Dfsu.plot()
> * DfsuLayered.read(): time_steps argument
> * DfsuLayered.get_layer_elements(): layer argument
> 
> What are your thoughts on these? Have you not seen these or have you decided we should postpone their retirement?
> 
> I think dfsu and dfs3 write methods should be consistent with the other write methods (and therefore should have these removed). It would also be good to get rid of the other deprecated functionality listed above IMO.

Most of these have now also been removed.

I think we can keep `Dfsu.plot()` a little longer.

I am unsure if the `Dfsu.find_nearest_elements()` has exactly the same functionality as `_GeometryFMLayered.find_index()`, I tried to adapt this test: https://github.com/DHI/mikeio/blob/2ba5981beacb88d5ec2f928a4d07150f7d91fba1/tests/test_dfsu_layered.py#L304 but I didn't quite get the same results🤔.

jsmariegaard:
> `Dfsu.find_nearest_elements()` has exactly the same functionality as `_GeometryFMLayered.find_index()`

find_nearest and find_index is *not* the same. find_index() will find the element containing a point. find_nearest_element() will find the element whose center is closest to a point.

jsmariegaard:
The intention was therefore not to remove find_nearest_element totally, but to remove it from the dfs class (and leave it on the geometry)

PR #419, opened 2022-09-08
updating pfs.py to do writing

clemenscremer:
updated pfs.py for writing capabilities with code originating from PFS_dict.py by PIBS

### DIFF
--- a/mikeio/pfs.py
+++ b/mikeio/pfs.py
@@ -3,6 +3,8 @@ import re
 import yaml
 import pandas as pd
 
+import os # TO BE DELETED LATER? 
+
 
 from types import SimpleNamespace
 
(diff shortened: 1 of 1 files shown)


mikeio/pfs.py
```diff
@@ -191,7 +191,7 @@ def catch_str_exceptions(self, v):
                     v = f"'{v}'"
 
         if isinstance(v, datetime):
-            #v = v.strftime("%Y, %#m, %#d, %#H, %M, %S") # pfs-datetime output
+            # v = v.strftime("%Y, %#m, %#d, %#H, %M, %S") # pfs-datetime output
             v = v.strftime("%Y, %-m, %-d, %-H, %-M, %-S")  # no zero padding
```
clemenscremer, line 195: this notation works only on unix-based systems

ecomodeller: I see, too bad.

ecomodeller:
We need an object model that can handle multiple roots like these files, which apparently are very common.
image

clemenscremer:
We need to be able to handle 
1. empty sections
2. duplicate parameters in one section (see type_of_radiation_stress_correction)

Screenshot showing both with original (left) and pfs.write output (right)
image

clemenscremer:
check that boolean output (e.g. include_sand parameter in the MT pfs) is lowercase. Currently it is written with capital first letter which doesn't seem to be mike standard

ecomodeller:
> We need to be able to handle
> 
> 1. empty sections
> 2. duplicate parameters in one section (see type_of_radiation_stress_correction)
> 
> Screenshot showing both with original (left) and pfs.write output (right) image

I don't understand how duplicate parameters in the same section can be valid. Do you?

jsmariegaard:
> I don't understand how duplicate parameters in the same section can be valid. Do you?

I don't think it should - bound to go wrong. 

I am "fixing" it right now

jsmariegaard:
We should stick to the official PFS naming convention (e.g. "target") as seen in https://docs.mikepoweredbydhi.com/core_libraries/pfs/pfs-file-content/ and https://manuals.mikepoweredbydhi.help/2021/General/Class_Library/DHI_PFS/html/N_DHI_PFS.htm

ecomodeller:
> We should stick to the official PFS naming convention (e.g. "target") as seen in https://docs.mikepoweredbydhi.com/core_libraries/pfs/pfs-file-content/ and https://manuals.mikepoweredbydhi.help/2021/General/Class_Library/DHI_PFS/html/N_DHI_PFS.htm

I agree that we should stick to the official standard.

And according to the docs there is no such thing as a PFS datetime [Parameter](https://manuals.mikepoweredbydhi.help/2021/General/Class_Library/DHI_PFS/html/T_DHI_PFS_PFSParameter.htm) ☹️

**Some examples:**

It could be a string formatted using ISO8601:
image

A string formatted in some other way:
image

Or a list of integers:
image

Also note the lack of convention for the keyword format:
* CREATEDTIME (UPPERCASE)
* StartTime (PascalCase)
* start_time (snake_case)

PR #424, opened 2022-09-12
Spectral subplots

ecomodeller:
Adressing


ecomodeller:
Comparing two timesteps is now possible, colorbars are however not identical by defalt;
image

Adding a colorbar is not great:
image

Maybe passing an `ax` as an argument is too low level and we should aim for a higher level of abstraction baked into the `.plot()` methods like [faceting in xarray](https://docs.xarray.dev/en/stable/user-guide/plotting.html#faceting)

PR #431, opened 2022-09-16
add label argument to dfs2 plots

miab-dhi:
Adapted the color bar parameters to include an label argument in the plot methods.


jsmariegaard:
Great, @miab-dhi .

I wonder if this was the only plot method that missed the label argument or if it is also needed elsewhere 🤔

miab-dhi:
>

ecomodeller (approved):

PR #435, opened 2022-09-30
DataArray/Dataset extract track

ecomodeller:


ecomodeller:
Could be that this needs more tests, I get one failing test in fmskill when I try to use this branch and have a track observation...

ecomodeller:
Note to self: 
This happens in the Multi_variable case:
image

image

mikeio/track.py
```diff
@@ -0,0 +1,158 @@
...
+
+        while (read_next == True) and (not is_EOF(dfsu_step + 1)):
+            dfsu_step = dfsu_step + 1
+
+            # swap new to old
+            d1, d2 = d2, d1
+            t1, t2 = t2, t1
+
+            step = time_steps[dfsu_step]
+            for item in range(n_items):
```
ecomodeller, line 126: Take a closer look here

ecomodeller:
* Should the method be called `Dataset.extract_track()` or should it be merged with the `Dataset.sel()` ?
* Should it return a Dataset with location as lon, lat items or should we create a new type of geometry?

PR #436, opened 2022-09-30
Pfs allow repeated keywords

### LINKED ISSUE #434: pfs to read and modify plc files
Dear all,

I am using the mikeio.read_pfs to read, modify and rewrite pfs files. While this works well with .dfsu files, it gives troubles with .plc files.

Simply reading a plc file and rewriting the same identical renamed file makes it lose the colour palette (warning: "RGB_Color_Value defined multiple times").
Under the section [IMAGE_FILES], the ImageFileName goes from |\\dkcph1-stor2.dhi.dk\project... to '|\\dkcph1-stor2.dhi.dk\project..', with these '' symbols resulting in altering the path. 
In addition, changing the [DataFile] path section deletes sections and makes the file invalid. 

Please find attached the comparison between the two .plc files (the read plc and the rewritten and renamed file). 
The line of codes is the following:

main_dir_frv = '//dkcph1-ncr3626/FRMA/Iteration_flushing_py/postprocess_YEAR/'
file_frv = 'AD2D_d37_channel_0_duration_d'
file_path_plc = os.path.join(main_dir_frv, file_frv + '.' + filename_suffix_plc)
pfs_plc = mikeio.read_pfs(file_path_plc)

new_file_path_plc_input = os.path.join('|.\Try.dfu|')
pfs_plc.MIKE_ZERO_PLOT_COMPOSER.ACTIVE_X_PLOT.PLOT_PROPERTIES.Files.DataFile = new_file_path_plc_input
new_file_path_plc = os.path.join(main_dir_frv + 'try.plc')
pfs_plc.write(new_file_path_plc)

image

Capture
Capture2

jsmariegaard:
Allow the user to select whether he/she wants to allow repeated keywords in the same PfsSection by introducing a new argument in read_pfs() `unique_keywords` by default True. Relevant for MIKE Zero Plot Composer where `unique_keywords` should be set to False. Closes #434 

Fixes also bug with repeated keywords when unique_keywords=True (as it was before).

### DIFF
--- a/mikeio/pfs.py
+++ b/mikeio/pfs.py
@@ -172,21 +172,6 @@ def parse_yaml_preserving_duplicates(src, unique_keywords=True):
     class PreserveDuplicatesLoader(yaml.loader.Loader):
         pass
 
-    # DOES NOT ALLOW DUPLICATES:
-    # def map_constructor_duplicates(loader, node, deep=False):
-    #     keys = [loader.construct_object(node, deep=deep) for node, _ in node.value]
-    #     vals = [loader.construct_object(node, deep=deep) for _, node in node.value]
-    #     key_count = Counter(keys)
-    #     data = {}
-    #     for key, val in zip(keys, vals):
-    #         if key_count[key] > 1:
-    #             if key not in data:
-    #                 data[key] = []
-    #             data[key].append(val)
-    #         else:
-    #             data[key] = val
-    #     return data
-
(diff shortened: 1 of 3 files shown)


clemenscremer (approved):

ecomodeller:
`map_constructor_duplicate` and `map_constructor_duplicate_sections` are almost identical and could benefit from using a common function, but it gets impossible to understand what is going on. Let's leave it like this.

PR #438, opened 2022-10-01
Short unit names

jsmariegaard:
nicer on plots - e.g. 'm^2*s/deg' instead of 'meter_pow_2_sec_per_deg'.

Should we also change display_name: 

    items:
      0:  Energy density <Wave energy density> (meter pow 2 sec per deg)


ecomodeller:
While we are at it... should we also provide a version with units using math expression?
image

jsmariegaard:
Yes maybe - I guess the difference would mostly be on degrees. Note that typesetting of units in LaTeX should not be standard math mode (italics) https://tex.stackexchange.com/questions/509939/problem-with-units-in-math-mode

ecomodeller:
I was thinking about units like these: https://github.com/DHI/mikeio/blob/70949a4eae059a338398a5baee03d9a284e333a3/mikeio/eum.py#L884

which could be translated to (ignore the italics)

$$ g / m^2 / d $$

which is ambiguous since it could be interpreted either as

$$ \frac{\frac{g}{m^2}}{d} = g m^{-2} d $$

or

$$ \frac{g}{m^2  d}  = g m^{-2} d^{-1}$$

where the second version is most likely the correct one.

ecomodeller:
@jsmariegaard should we merge this, or do you have anything else to add?

jsmariegaard:
@ecomodeller you can merge 👍

PR #439, opened 2022-10-03
Pfs tab vs spaces

jsmariegaard:
Allow tabs in PFS files.


ecomodeller (approved):
Based on a quick survey of pfs files, tabs seem to be very rare.

And whitespace other than newlines seem to be ignored in general.

jsmariegaard:
> Based on a quick survey of pfs files, tabs seem to be very rare.

Yes, agree, that's also why we didn't notice before now. But a lot of people hand-edit these files - and if MIKE can read the files so should MIKE IO.

PR #441, opened 2022-10-03
dataset concatenation single timestep

clemenscremer:
modified dataset test to concatenate file with single timestep. When concantenating will result in error in shape check as well as dims check as time dimension is dropped for single timestep datasets and not available for checking.

### DIFF
--- a/mikeio/dataset.py
+++ b/mikeio/dataset.py
@@ -1243,9 +1243,9 @@ class Dataset(DataUtilsMixin, TimeSeries, collections.abc.MutableMapping):
 
     def _concat_time(self, other, copy=True) -> "Dataset":
         self._check_all_items_match(other)
-        if not np.all(self.shape[1:] == other.shape[1:]):
+        if not np.all(self.shape[1:] == other.shape[1:]): # CHECK DOES NOT WORK FOR SINGLE TIMESTEPS AS TIME DIM IS DROPPED
             raise ValueError("Shape of the datasets must match (except time dimension)")
-        if "time" not in self.dims:
+        if "time" not in self.dims: # DIM CHECK DOES NOT WORK. USE ATTRIBUTE CHECK INSTEAD?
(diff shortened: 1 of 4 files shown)


clemenscremer:
should we keep dimensions when selecting single timestep?
this would prevent checks during concatenation from failing

image

ecomodeller:
> should we keep dimensions when selecting single timestep? this would prevent checks during concatenation from failing
> 
> image

I don't think so.

We changed this behaviour to v 1.0 in order to make it consistent with e.g. Xarray and also to make plotting easier.

If you select a single timestep from a dataset with 3 dimensions (time, y, x) you end up with a dataset with two dimensions (y,x).

This is also how indexing in NumPy works
```python
>>> A
array([[[0.9, 0.4, 0.6, 0.4],
        [0.5, 0.4, 0.1, 0.4],
        [0.9, 0.4, 0.3, 0.9]],

       [[0.3, 0.7, 0.4, 0.8],
        [0.7, 0.7, 0.3, 0.4],
        [0.2, 0.8, 0.9, 0.8]]])
>>> A[:,:,0]
array([[0.9, 0.5, 0.9],
       [0.3, 0.7, 0.2]])
>>> A.shape
(2, 3, 4)
>>> A[:,:,0].shape
(2, 3)
```

But concatenation should still be able to work somehow 😉

jsmariegaard:
@clemenscremer - could you change the name of this PR to something more meaningful? Thanks

ecomodeller:
1. Make sure it works for most (all?) sorts of data.
2. Is it consistent with `generic.concat`?
3. Release as part of v 1.2 🚀

jsmariegaard (approved):
Looks good. I tried hard to break with all sorts of dfs files, combinations of single and multi-timesteps for both Dataset and DataArray. But I couldn't. Well done!

PR #442, opened 2022-10-04
Pfs mixed array

### LINKED ISSUE #437: PFS. read write of parameter list, which contains both boolean and string
MIKE IO version 1.2.dev0

This is probably a rare case, but it seems that this line will fail: 

    fill_list = false, 'TEST'

if you read it from a pfs file and try to write it again you will get this: 

    fill_list = 'false, 'TEST''

Which is different. This only occurs if the line contains both boolean and string.  

Here is another example: 

    fill_list = 0,'TEST',false

which gives

    fill_list = '0,'TEST',false'

Some other combined cases can be even worse(!) - the below line cannot be parsed: 

     fill_list = 'dsd', 0, 0.0, false

Gives this error:

      E           expected <block end>, but found ','
      E             in "<unicode string>", line 16, column 22:
      E                   fill_list:  'dsd', 0.0, false

Probably because the string is first on the line.

ecomodeller:
Closes #437


mikeio/pfs.py
```diff
@@ -453,7 +456,9 @@ def _parse_line(self, line: str) -> str:
                 key = key.strip()
                 value = s[(idx + 1) :]
 
-                if s.count("'") == 2:  # This is a quoted string and not a list
+                if (
+                    s[0] == "'" and s[-1] == "'"
```
jsmariegaard, line 460: keyword = "str1", "str2"

    keyword = "str1", 0.0, false, "str2"

I guess you still need the `s.count("'") == 2` condition?

ecomodeller: Go ahead and add a failing test 😉

ecomodeller: 🚀 science

jsmariegaard:
:shipit: ?

Issue #446, opened 2022-10-12
Reading a pfs file with embedded string, e.g. 'hello ''to'' the world'


ecomodeller:
**Describe the bug**
Reading a pfs file with embedded strings changes the formatting to a format not accepted by Mike Zero

**To Reproduce**
Example text:

```
[ECOLAB_MODULE]
    [DERIVED_VARIABLES]
        [DERIVED_VARIABLE_106]
            Touched = 0
            name = 'alfa_PC_T'
            type = 0
            dimension = 3
            description = 'alfa_PC_T, ''light'' adjusted alfa_PC, ugC/gC*m2/uE'
            EUM_type = 999
            EUM_unit = 0
            unit = 'ugC/gC*m2/uE'
            ID = 597
        EndSect  // DERIVED_VARIABLE_106
    EndSect
EndSect
```

```python
>>> import mikeio
>>> mikeio.__version__
'1.2.0'
>>> pfs = mikeio.read_pfs("tiny.m3fm")
>>> pfs.write("tiny_out.m3fm")
```

**Expected behavior**
Intact quoting of embedded string

**Screenshots**
![image](https://user-images.githubusercontent.com/614215/195353765-65ec268e-b43d-43a4-a71f-43d0258bb6f5.png)

**System information:**
 - Python 3.10.6
 - MIKE IO version 1.2.0

ecomodeller:
Here is another tricky example: see EXPRESSION line
```
      [AUXILIARY_VARIABLE_7]
         SYMBOL = 'Hrms'
         ID = 422
         DESCRIPTION = 'Hrms, root-mean-square wave height (m) (=Hs/sqrt(2))'
         ONLINE_HELP = ''
         DOCUMENTATION = ''
         EUM_TYPE = 999
         EUM_UNIT = 0
         OUTPUT = 'YES'
         SCOPE = 'NOT_SPECIFIED'
         SPATIAL_VARIATION = 'HORISONTAL'
         UNIT = 'm'
         EXPRESSION = '"IF(depth>0.0,IF(Hs>0.0,Hs/SQRT(2.0),0.0),0.0)"'
      EndSect  // AUXILIARY_VARIABLE_7
```

ecomodeller:
Fixed by #448

PR #448, opened 2022-10-13
Pfs quotes

ecomodeller:
Closing #446


clemenscremer (approved):
Works like a charm. Could we - in theory also experience double doublequotes (""value"") and should we adress it as well?

ecomodeller:
> Works like a charm. Could we - in theory also experience double doublequotes (""value"") and should we adress it as well?

Let's deal with it when theory becomes practise.

PR #450, opened 2022-10-25
Update full_test.yml

ecomodeller:


.github/workflows/full_test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10","3.11"]
```
ecomodeller, line 21:
```suggestion
        python-version: [3.7, 3.8, 3.9, "3.10","3.11.0-rc.2"]
```

jsmariegaard:
E.g. also on documentation: https://dhi.github.io/mikeio/#requirements

ecomodeller:
> I think it would be great to show that we support 3.11 in the png on the readme, in setup etc...

Fixed in this PR https://github.com/DHI/mikeio/pull/453
Seemed like overkill to create a new PR only for this...

PR #451, opened 2022-10-26
Make scipy optional

ecomodeller:
Scipy is currently the largest dependancy of MIKE IO (in MB) and is *only* used for:
* Interpolation in time
* Search trees

Thus, I suggest to make it optional similar to how xarray stays slim and uses [optional dependencies](https://docs.xarray.dev/en/stable/getting-started-guide/installing.html#optional-dependencies).

$ pip install mikeio

Downloading mikeio-1.2.1-py3-none-any.whl (196 kB)
Downloading mikecore-0.2.1-py3-none-any.whl (17.1 MB)
Downloading numpy-1.23.4-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (17.1 MB)
Downloading PyYAML-6.0-cp310-cp310-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_12_x86_64.manylinux2010_x86_64.whl (682 kB)
Downloading xarray-2022.10.0-py3-none-any.whl (947 kB)
Downloading scipy-1.9.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (**33.7 MB**)
Downloading tqdm-4.64.1-py2.py3-none-any.whl (78 kB)
Downloading pandas-1.5.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (12.1 MB)
...

### DIFF
--- a/mikeio/dfsu_layered.py
+++ b/mikeio/dfsu_layered.py
@@ -3,7 +3,6 @@ import pandas as pd
 import warnings
 from tqdm import trange
 from functools import wraps
-from scipy.spatial import cKDTree
 
 from mikecore.DfsuFile import DfsuFile, DfsuFileType
 from .dfsu import _Dfsu
@@ -391,6 +390,9 @@ class Dfsu3D(DfsuLayered):
         --------
         >>> dfsu.extract_surface_elevation_from_3d('ex_surf.dfsu')
         """
+
+        from scipy.spatial import cKDTree
+       
         # validate input
         assert (
             self._type == DfsuFileType.Dfsu3DSigma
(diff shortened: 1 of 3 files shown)


jsmariegaard:
should we make tqdm and xarray optional as well? I guess they are not needed for most applications....

ecomodeller:
> should we make tqdm and xarray optional as well? I guess they are not needed for most applications....

True, but they are small and fast to install. So the benefit is not very big, unless they start to have conflicting dependencies.

jsmariegaard:
Yes, but I was thinking that each package has a baseline plus an MB cost. And these packages might have dependencies too...

ecomodeller:
It seems like the benefit of making it optional is not big enough at the moment to motivate this. Closing this for now.

PR #453, opened 2022-10-26
Faster dfs{2,3} write

ecomodeller:
Before:
image

After:
image


ecomodeller:
Thanks to @PierreBrinkSwiegers for pointing out this flaw 👏

jsmariegaard (changes requested):
Could you please fix for dfs3 as well https://github.com/DHI/mikeio/blob/5a9427168b39efab5c96de8ccff9b7b87f444932/mikeio/dfs3.py#L114 ?

ecomodeller:
> Could you please fix for dfs3 as well
> 
> https://github.com/DHI/mikeio/blob/5a9427168b39efab5c96de8ccff9b7b87f444932/mikeio/dfs3.py#L114
> 
> ?

Hmm, looks like somebody has been copy/pasting their way to productivity instead of writing clean code. Who could it be?

```console
$ git blame mikeio/dfs3.py|grep "def _write_dfs3_data"
5b98457a mikeio/dfs3.py (Henrik Andersson          2022-04-20 09:35:07 +0200  97) def _write_dfs3_data(dfs: DfsFile, ds: Dataset) -> None:
```

Time for some light refactoring :see_no_evil:

jsmariegaard:
I was also wondering if this is the best way to find if a DatetimeIndex is equidistant: 

    return len(self.time.to_series().diff().dropna().unique()) == 1 

? 

And/or whether we should cache this?

ecomodeller:
> I was also wondering if this is the best way to find if a DatetimeIndex is equidistant:
> 
> ```
> return len(self.time.to_series().diff().dropna().unique()) == 1 
> ```
> 
> ?
> 
> And/or whether we should cache this?

It should be safe to cache this, the time axis should be immutable.

PR #458, opened 2022-10-28
add dfs0 property items_dataframe

Hendrik1987:
@ecomodeller , I often find myself needing to get an overview of the content of dfs files without reading any data. Returning this as a dafarame allows easy export and stacking when looping multiple files. For now I added to dfs0 only and you can do:

```
import mikeio
ds = mikeio.open("testdata/random.dfs0")
ds.items_dataframe
```
image

Seems like it would have to be added to many places. And maybe it's just a (too) specialized form of `ds.items`?

### DIFF
--- a/mikeio/dfs0.py
+++ b/mikeio/dfs0.py
@@ -482,20 +482,6 @@ class Dfs0(TimeSeries):
         return self._items
 
     @property
-    def items_dataframe(self):
-        "pd.DataFrame of items"
-        df = pd.DataFrame(
-            data = dict(
-                name = [i.name for i in self._items],
-                EUMUnitInt = [i.unit for i in self._items],
-                EUMUnit = [i.unit.display_name for i in self._items],
-                EUMTypeInt = [i.type for i in self._items],
-                EUMType = [i.type.display_name for i in self._items],
-            )
-        )
-        return df
-
-    @property
     def start_time(self):
         """File start time"""
         return self._start_time
--- a/mikeio/dfs0.py
+++ b/mikeio/dfs0.py
@@ -482,6 +482,20 @@ class Dfs0(TimeSeries):
         return self._items
 
     @property
+    def items_dataframe(self):
+        "pd.DataFrame of items"
+        df = pd.DataFrame(
+            data = dict(
+                name = [i.name for i in self._items],
+                EUMUnitInt = [i.unit for i in self._items],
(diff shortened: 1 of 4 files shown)


ecomodeller:
Sure, similar to the item information table you get in Mike Zero:
image

Why is it necessary to include the Int version of EUMType and EUMUnit, the fact that they are stored as integers, should preferably not be necessary to care about?

ecomodeller:
@Hendrik1987 I created an `ItemInfoList` with a `to_dataframe()` method. I think this is cleaner but still the same purpose.

Let's consider what to put in the dataframe.

Hendrik1987:
You are right, integers are probably not needed. I thought if one wanted to use this as input for ItemInfo, though I cannot think of a use case.
The mike zero table is a good template, though stats are of course only possible if data is loaded, which is then already covered by describe.
Btw, with mike zero I can read and write on the Pos(x, y, z) attr. So it must be stored in the dfs0 file somewhere. Can that field be accessed through mikeio?

ecomodeller:
> You are right, integers are probably not needed. I thought if one wanted to use this as input for ItemInfo, though I cannot think of a use case. The mike zero table is a good template, though stats are of course only possible if data is loaded, which is then already covered by describe. Btw, with mike zero I can read and write on the Pos(x, y, z) attr. So it must be stored in the dfs0 file somewhere. Can that field be accessed through mikeio?

I think the global item statistics can actually be read [without reading all the data](<https://docs.mikepoweredbydhi.com/core_libraries/dfs/dfs-file-system/#statistics-stored-in-the-dfs-file>). @JesperGr can we access these data (min, max, mean) in mikecore-python?

Hendrik1987:
I think the column names should also be aligned with the names of the attributes here: https://github.com/DHI/mikeio/blob/ec9cbd975e9ffe06fbad5937507fcf13ee18cf11/mikeio/dataarray.py#L2255

However, I like `item.type.name` and `item.info.name` better as done in this PR.

Though, for the xarray ds I find `name` as a key of the attrs dict maybe a bit confusing as now there is a `da.name` and `da.attrs.name` (which can be different). So maybe we can consider `item_name` both for the attrs and the items table.

Hendrik1987:
@ecomodeller : Any chance we can settle on this and close soon? Suggest to do without the stats for now and open that part as an enhancement. Thanks!

ecomodeller (approved):

PR #460, opened 2022-11-01
Fix dfs3 single layer

jsmariegaard:
Before this PR it was not possible to read single-layer dfs3 files with MIKE IO. This PR also enables the `keepdims` keyword for reading dfs3s - it was previously inactive and therefore inconsistent with all other dfs read methods.


jsmariegaard:
Closes #461

ecomodeller (approved):

Issue #463, opened 2022-11-02
MIKE.she pfs files cannot be parsed


watermain:
Try to load any MIKE SHE pfs file (.she, just create a new doc in MZ with no need to make any settings) with mikeio.read_pfs
Errors:
1. AttributeError: 'float' object has no attribute 'append'
2. During handling of the above exception, another exception occurred

Python version: 3.11
MIKE IO version: 1.2.2

jsmariegaard:
Thanks @watermain for testing and reporting. Much appreciated. Yesterday I ran a test and can confirm the problem. If you don't require unique_keywords you can parse it with: 

    > pfs = mikeio.read_pfs("Karup_basic.she", unique_keywords=False) 

Unfortunately, this doesn't really solve the problem. If you write the file again to disk (with pfs.write() ) and compare the original and new file, three problems show: 

* If a section has keywords and subsections with the same name, the subsection is not parsed correctly (it is considered as a standard dict instead of a PfsSection). This is apparently common practice in MIKE SHE. We need to be able to handle not just repeated keywords and repeated sections but also combinations of these. 
* Strings that are a number e.g. '1' will be converted to a float upon _writing_. They are parsed correctly, and the problem first occurs at the time of writing. But obviously, need to be fixed anyway. 
* Lists containing vertical lines are not written correctly 

Original: 
    TemplateFile_OL = ||, -1, -1, -1, -1, -1, -1

Written by MIKE IO
    TemplateFile_OL = '||', -1, -1, -1, -1, -1, -1

jsmariegaard:
And one thing more that I missed first: this projection string is not written correctly back to file either: 

Original:

      MapProjShp = 'PROJCS["ETRS_1989_UTM_Zone_32N",GEOGCS["GCS_ETRS_1989",DATUM["D_ETRS_1989",SPHEROID["GRS_1980",6378137.0,298.257222101]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Transverse_Mercator"],PARAMETER["False_Easting",500000.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",9.0],PARAMETER["Scale_Factor",0.9996],PARAMETER["Latitude_Of_Origin",0.0],UNIT["Meter",1.0]]'

Written by MIKE IO: 

[...]

PR #464, opened 2022-11-02
Parse MIKE SHE Pfs

ecomodeller:


ecomodeller:
@jsmariegaard The parsing error occurs inside the `parse_yaml_preserving_duplicate`, but it seems it can be parsed with the normal loader, should this be the default strategy to try first instead of last, I can't really tell when to use which loader.

jsmariegaard:
@ecomodeller, please see my analysis in https://github.com/DHI/mikeio/issues/463

jsmariegaard:
Furthermore, it would be nice if the parse could provide some feedback on where in the file it failed...

jsmariegaard:
@ecomodeller, note that I changed the default to NOT insist on unique keywords/section names. This may not be ideal for MIKE FM type of models but as we have seen for MIKE SHE and other engines this is the standard. We could have `None` as the default value and then have a list of file extensions to dynamically set the default behavior, but I don't know... what do you think?

jsmariegaard:
Also, the implementation is now (after handling combinations of non-unique keywords and sections) less elegant as can be seen from the diff, so ideas on improving is very welcome.

jsmariegaard:
And... I am not sure if we really need the long complete MIKE SHE file, now that we have the smaller condensed one?

ecomodeller:
> And... I am not sure if we really need the long complete MIKE SHE file, now that we have the smaller condensed one? 

Probably not.

ecomodeller:
@watermain can you test if this branch works with some of your MIKE SHE examples?

The branch can be installed like this:
```console
$ pip install https://github.com/DHI//mikeio/archive/pfs_she.zip
```

watermain:
Hi @ecomodeller, I think there must be an installation issue: The problems persists after updating as suggested. Any ideas?

ecomodeller:
I guess you have to add --upgrade --force-reinstall, since you already have mikeio installed.

watermain:
Ok cool, reading seems to work now! Writing however fails - generally, it seems, so I´m not adding a sample she file, but let me know if you need more details!

jsmariegaard:
> Ok cool, reading seems to work now! Writing however fails - generally, it seems, so I´m not adding a sample she file, but let me know if you need more details!

@watermain , could you please provide a minimal failing test to this PR?

watermain:
With the latest commit read/write of she files works in almost all cases. I have run tests on many old support cases, and there are a few remaining special cases: Apostrophes in filenames don´t work, like this one:
FILE_NAME = |.\Inputs\11_Storing of results\ouvrages\Pont de l'Angle 1h.dfs0|
Strings that could almost be parsed as floats, but should still be just handled as strings, like this:
ItemDescr = 'E-1'
(or 'E12'...)
Then there is one file where it does not say which pfs line makes it fail, so I added the whole file DK4_100m_2020_Ribe_Submodel_flood_run-70ems3b.she to test data, it´s a bit large though, so feel free to discard it before merging. I added the previous 2 cases to this file, so the first 2 issues may need to be fixed to see the 3rd issue.
I also included other pfs types in the test, and there may be more issues, but I´ll leave it at this for now.

ecomodeller:
Writing non-unique keywords fails in this branch, but works in main. The scientific approach to finding where the regression was introduced is bisection using [`git bisect`](https://git-scm.com/docs/git-bisect)

This is the result:
```console
759a4cc580cdcf80f66635fcc8f939d3ae71a395 is the first bad commit
commit 759a4cc580cdcf80f66635fcc8f939d3ae71a395
Author: Jesper Sandvig Mariegaard <34088801+jsmariegaard@users.noreply.github.com>
Date:   Thu Nov 3 16:08:47 2022 +0100

    handle non-unique mix of keywords and sections

 mikeio/pfs.py     | 24 ++++++++++++++++++------
 tests/test_pfs.py |  2 +-
 2 files changed, 19 insertions(+), 7 deletions(-)
```

jsmariegaard:
@watermain I fixed all known issues. Could you please confirm that it now works? 

It should be noted, though, that certain combinations of characters in strings can still confuse it, e.g.: 

     E = |str,s'+-s_d.dfs0|

See the disabled lines in test_difficult_chars_in_str().  https://github.com/DHI/mikeio/blob/97f4149e7bd43e68e3932aacfc0055f1c6b62846/tests/test_pfs.py#L602 

@ecomodeller or @watermain could one of you please review the PR a final time? I think we should then merge.

jsmariegaard:
Thanks @watermain, it's clear that we need better parsing of parameters. Maybe we should use csv or shlex module to split the strings in tokens.

jsmariegaard:
@watermain and @ecomodeller - I think we should merge this PR now even though a few issues remain unsolved. I would propose creating those as issues and then opening a new PR. @ecomodeller also suggested having an option to allow loading a pfs file in text mode which could be a fallback strategy if the normal parsing fails.

jsmariegaard:
closes #470

jsmariegaard:
All known issues have been fixed except strings that could also be a valid scientific float representation e.g. 

    ID_A = '1E-3'

We will need to make this an issue and work on it in another PR.

ecomodeller (commented):
The test coverage for pfs.py is 87 %

Most lines missing tests are not created in this PR. But it would still be good to know if the untested lines are actually used.

For example this method:
image

PR #469, opened 2022-11-16
fix a bug in reading and writing pfs with duplicate keywords

mohm-dhi:
There was a case that there was two keywords that one only had a value and the second one was a PfsSection. In this situation the read pfs function reads the second occurrence as a dictionary and it writes back to file as dictionary again. That is been fixed here.

### DIFF
--- a/mikeio/pfs.py
+++ b/mikeio/pfs.py
@@ -78,8 +78,8 @@ class PfsSection(SimpleNamespace):
             self.__setattr__(key, value)
 
     def pop(self, key, *args):
-        """If key is in the dictionary, remove it and return its 
-        value, else return default. If default is not given and 
+        """If key is in the dictionary, remove it and return its
+        value, else return default. If default is not given and
         key is not in the dictionary, a KeyError is raised."""
         return self.__dict__.pop(key, *args)
 
@@ -223,7 +223,7 @@ def parse_yaml_preserving_duplicates(src, unique_keywords=True):
                 if isinstance(val, dict):
                     if key not in data:
                         data[key] = []
-                    data[key].append(val)
+                    data[key].append(PfsSection(val))
                 else:
                     if key not in data:
                         data[key] = PfsRepeatedKeywordParams()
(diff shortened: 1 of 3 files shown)


jsmariegaard:
@mohm-dhi please be aware of #464

mohm-dhi:
@jsmariegaard I moved the issue to the issue section that I thing it makes confusion about the topic of this PR, but as I mentioned I fixed a bug related to reading duplicate keywords when one has value but the other one is PfsSection. You can try with the current version to repeat the bug.

mohm-dhi:
I can see the above issue is fixed now but another issue introduced that was not there before, see screenshot below, you can see duplicate keywords in separated lines now are merged together in one line separated with comma. 


image

jsmariegaard:
This should not be the case on the main branch. Could you please check?

mohm-dhi:
> This should not be the case on the main branch. Could you please check?

I tested that in the main branch, I just double checked as well.

jsmariegaard:
Okay - I will do a bit of testing myself to see if I can reproduce. Thanks for reporting :-)

mohm-dhi:
> Okay - I will do a bit of testing myself to see if I can reproduce. Thanks for reporting :-)

alright, here is the file I use if you want to reproduce it. 
[NARW_testing_02_single_particle_test.txt](https://github.com/DHI/mikeio/files/10045053/NARW_testing_02_single_particle_test.txt)

just change the extension to .mfm

and this is my code: 

    file = r"NARW_testing_02_single_particle_test.mfm"
    pfs = mikeio.Pfs(file)
    pfs.write(file[:-5] + "_new.mfm")

jsmariegaard:
Thanks, I can confirm there is an issue here. Could you please report it as a new issue? - I will try to fix it over the weekend.

jsmariegaard:
Issue created #473 and fixed in https://github.com/DHI/mikeio/pull/474

ecomodeller:
@mohm-dhi is this PR still relevant?

mohm-dhi:
> @mohm-dhi is this PR still relevant?

no you can close it.

PR #477, opened 2022-11-19
Pfs refactor write

jsmariegaard:
Move internal write method to PfsSection such that it can be used for __repr__. This makes sure that the __repr__ etc look exactly like what is written to file. This will also make it easy to write to a stream instead of a file.

### DIFF
--- a/mikeio/pfs.py
+++ b/mikeio/pfs.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 from types import SimpleNamespace
-from typing import List, Tuple
+from typing import List, Tuple, Callable
 from collections import Counter
 from datetime import datetime
 import re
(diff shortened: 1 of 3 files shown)


ecomodeller:
I am not sure about the `.head()` and `.tail()` methods. It is not consistent with pandas.Dataframe.head (inspired by unix head).

The pandas version takes a Dataframe and selects a subset, which is also a Dataframe. Same with unix head, it selects a subset of lines of text, which allows you to either print it to the termnal or pipe it to some other program.

When is it useful, to get a subset of a PfsSection? Can you come up with another name that does the job without the risk of causing confusion?

ecomodeller:
How about we skip `head` and `tail` for now but merge the other stuff?

jsmariegaard:
head() and tail() removed

ecomodeller (approved):

PR #478, opened 2022-11-20
Pfs. Find keys and params

jsmariegaard:
image

image

image


jsmariegaard:
Could perhaps benefit from a bit more testing of search in cases with repeated keywords and/or sections...

ecomodeller:
Is this the behavior we want?
image

It seems a bit funny that the returned result is of type `PfsSection` but has no section header?

PR #483, opened 2022-11-24
Xarray point add missing coordinates for GeometryPoint{2,3}D

ecomodeller:


mikeio/dataarray.py
```diff
@@ -2245,6 +2245,13 @@ def to_xarray(self):
             coords["element"] = xr.DataArray(
                 data=self.geometry.element_ids, dims="element"
             )
+        elif isinstance(self.geometry, GeometryPoint2D):
+            coords["x"] = self.geometry.x
```
Hendrik1987, line 2249: Any attributes to add from `geometry`, e.g., info on projection and units? Actually same for the other geometries here. But `GeometryPoint2D.projection` does not seem to be defined.

Hendrik1987 (approved):
Yes, that is a good start! Only 1 minor thought for future PRs.

Hendrik1987:
Would it maybe be cleaner to add `.to_xarray_coordinates` to the respective `geometry` classes, rather than having all the if statements [here](https://github.com/DHI/mikeio/blob/c977af067c9f6361682efa9d13aa8317c3e5d2eb/mikeio/dataarray.py#L2235)? Here it could simply be after defining time: `coords.update(self.geometry.to_xarray_coordinates)`.
Pros and cons, just a thought...

PR #488, opened 2022-11-30
Nested PfsSection should have a parent

ecomodeller:
Needs more work, when you introduce attributes that are not part of the data structure, this doesn't work anymore
`
for k, v in vars(self).items():
` 
or
`d = self.__dict__.copy()
`
since `self.__dict__` now includes a `_parent` attribute which should be ignored in serialization.


jsmariegaard:
Can you remove _parent from:

def keys(self):
        """Return a new view of the PfsSection's keys"""
        return self.__dict__.keys()

and maybe also items() ???

PR #492, opened 2022-12-01
Fix interpolation of dfs1

ecomodeller:


ecomodeller:
Extrapolation now behaves exactly like `xarray`, since we use `xarray`🙄

ecomodeller:
Start from scratch in new PR

PR #495, opened 2022-12-05
Fix 1d interpolation

ecomodeller:
Interpolation in space for grid1d was previously only correct if the domain was [0,1].


jsmariegaard (approved):
Works great. Added a few minor changes.

PR #501, opened 2022-12-08
Read corrupt dfs files

ecomodeller:
Reading a corrupt timestep apparently messes up some internals.

A possible solution could be to close the dfs object and re-open it.

Not sure how to QA this PR☹️ But at least it should it not break any existing functionality.

### DIFF
--- a/mikeio/__init__.py
+++ b/mikeio/__init__.py
@@ -71,6 +71,11 @@ def read(filename, *, items=None, time=None, keepdims=False, **kwargs) -> Datase
     layers: int, str or sequence, optional
         Dfs3/Dfsu-layered: read only data from specific layers,
         by default None (=all layers)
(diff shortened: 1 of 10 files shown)


ecomodeller:
image

ecomodeller:
image
Gaps seems plausible.

ecomodeller:
Default behaviour is to raise error on bad data. Opt in to ignore bad data and the ability to choose a value to fill gaps (default missing value)
image

ecomodeller:
fill_corrupt

jsmariegaard (approved):
I think it would be best to have the same functionality for dfs123, but I guess that can come later. 
It would be good to add the new arguments to `__init__.py/read()` docstring. Ideally, a small example in `interp_na()` docstring would be nice. Otherwise good.

Issue #503, opened 2022-12-12
Dataset/DataArray aggregation does not work if no time axis 


jsmariegaard:
If a DataArray has no time axis, e.g.: 

![image](https://user-images.githubusercontent.com/34088801/207096407-09f46ebd-3ff0-4c69-8a9e-2345b6aef2d0.png)

The aggregation methods such as da.mean() will fail --- even if one tries to specify an axis argument. The same is the case for Datasets.

PR #508, opened 2022-12-14
Fix aggregate with no time axis

jsmariegaard:
Dataset/DataArray aggregation methods such as .mean() does not work if the DataArray has no time axis.


jsmariegaard:
Closes #503

PR #514, opened 2022-12-20
Speed up some slow tests

ecomodeller:


jsmariegaard:
Good idea - I have been thinking about we should do many of the layered dfsu test only on subset of the sigma-z file which is now 1.5MB.

PR #517, opened 2022-12-23
Winter cleaning

ecomodeller:
* Remove base class -> use PfsSection instead
* Split classes into multiple files


ecomodeller:
```mermaid
classDiagram
    class MutableMapping
    class SimpleNameSpace
    class PfsSection{
   +__getitem__(key)
  +__settitem__(key, value)
   }

    class PfsDocument{
   +List~PfsSection~ targets 
    +write(filename)
    }
    SimpleNameSpace <|-- PfsSection
    MutableMapping <|-- PfsSection
    PfsSection <|-- PfsDocument
```

PR #522, opened 2023-01-10
Geometry refactoring

ecomodeller:
The base class `_Geometry` had a bit too many functions not used in point or 1d geometries.


ecomodeller:
- Todo: `find_index()` how to handle points outside the domain

ecomodeller:
> * Todo: `find_index()` how to handle points outside the domain

In another PR

mikeio/spatial/FM_geometry.py
```diff
@@ -476,6 +523,13 @@ def type_name(self):
         """Type name, e.g. Mesh, Dfsu2D"""
         return self._type.name if self._type else "Mesh"
 
+    @property
+    def ndim(self) -> int:
```
ecomodeller, line 527: @jsmariegaard Is there more to it than this?

mikeio/spatial/FM_geometry.py
```diff
@@ -768,7 +824,9 @@ def _find_element_2d(self, coords: np.array):
                     coords[k, 0],
                     coords[k, 1],
                 )
-                ids[k] = many_nearest[lid] if lid > 0 else -1
+                ids[k] = (
+                    many_nearest[lid] if lid > 0 else -1
```
jsmariegaard, line 828: We need to fix this, see #539

mikeio/spatial/grid_geometry.py
```diff
@@ -635,8 +644,12 @@ def _xy_to_index(self, xy):
...
         x = xy[:, 0]
 
-        ii = (-1) * np.ones_like(x, dtype=int)
-        jj = (-1) * np.ones_like(y, dtype=int)
+        ii = (-1) * np.ones_like(
+            x, dtype=int
+        )  # TODO -1 is not a good choice, since it is a valid index
+        jj = (-1) * np.ones_like(
+            y, dtype=int
+        )  # TODO -1 is not a good choice, since it is a valid index
```
jsmariegaard, line 652: Should we replace all the -1 with -999999999 for now? It's certainly not optimal - but it will not be a valid index...

mikeio/spatial/FM_geometry.py
jsmariegaard: Are you referring to the PR in general or something specific?

ecomodeller: Only referring to the `ndim` property

PR #523, opened 2023-01-10
No private testing

ecomodeller:
Private methods and fields should not be tested, since they should be treated as an implementation detail not to depend upon.

Solution, remove some tests and add some public property to make this explicit.


ecomodeller:
Some inspiration on how to structure public, internal and private stuff can be found in sklearn e.g the neighbors module https://github.com/scikit-learn/scikit-learn/tree/main/sklearn/neighbors
where all submodules all prefixed with underscore, and only the classes that should be exposed to the end user is imported in the module namespace in  `__init__.py`
image

image

Using this strategy we can the test the public api of internal classes, but also signal to end-users that they should not import a submodule starting with an underscore.

mikeio/dataarray.py
```diff
@@ -1021,6 +1021,10 @@ def values(self, value):
 
         self._values = value
 
+    @property
```
ecomodeller, line 1024: Add this only the fly instead?

PR #527, opened 2023-01-10
Store all attributes of GeometryFM in Xarray

ecomodeller:
```mermaid
flowchart LR
    Dfsu <--> mikeio.Dataset <--> xarray.Dataset <--> NetCDF
    xarray.Dataset <--> Zarr
```


ecomodeller:
An alternative worth considering is to add support for [Xugrid](https://deltares.github.io/xugrid/) supporting the [UGRID conventions](https://ugrid-conventions.github.io/ugrid-conventions/)

PR #528, opened 2023-01-11
_GeometryFMLayered always has n_layers

ecomodeller:
Which eliminates some checks


ecomodeller:
Exposed as  sub-classes `GeometryFM3D`, `GeometryFMVerticalProfile`

PR #529, opened 2023-01-11
Interpolation in 3d not yet supported

ecomodeller:


jsmariegaard (changes requested):
Shouldn't this be added on the corresponding interp() method on DataArray?

ecomodeller:
> Shouldn't this be added on the corresponding interp() method on DataArray?

Of course, I 'll move it there. No need to have the check twice.

PR #531, opened 2023-01-11
Test downstream package (fmskill)

ecomodeller:


.github/workflows/downstream_test.yml
```diff
@@ -0,0 +1,33 @@
+name: Downstream test
+
+on:
+  push:
+    branches: [ main]
+  pull_request:
```
jsmariegaard, line 6: Should we remove testing on PR and only keep it when pushing to main as this is a secondary concern? In this way, we would still catch before releasing.

jsmariegaard (approved):

PR #532, opened 2023-01-12
DRY mikeio.open

ecomodeller:


mikeio/__init__.py
```diff
@@ -150,25 +150,23 @@ def open(filename: str, **kwargs):
...
+        "dfs2": Dfs2,
+        "dfs3": Dfs3,
+        "dfsu": Dfsu,
+        "mesh": Mesh,
+    }
+
+    if file_format not in READERS:
+        valid_formats = ", ".join(READERS.keys())
+        raise Exception(
+            f"{file_format} is not a supported format for mikeio.open. Valid formats are {valid_formats}"
```
jsmariegaard, line 167: Maybe add hint about mikeio.read_pfs for MIKE configuration files in the PFS format?

mikeio/__init__.py
```diff
@@ -150,25 +150,23 @@ def open(filename: str, **kwargs):
...
+        )
+
+    reader_klass = READERS[file_format]
```
jsmariegaard, line 170: klass? class?

ecomodeller: I first named it class which is a reserved keyword, then renamed to klass, then added. reader_

I'll rename it again.

jsmariegaard:
@PaulDanielML made me aware that this method does not take Path (only str). I think it should support Path - and then maybe add type hint in function signature.

PaulDanielML:
> @PaulDanielML made me aware that this method does not take Path (only str). I think it should support Path - and then maybe add type hint in function signature.

@jsmariegaard if you like I can add those things to the PR (not sure if I have push rights for mikeio though)

jsmariegaard:
> > @PaulDanielML made me aware that this method does not take Path (only str). I think it should support Path - and then maybe add type hint in function signature.
> 
> @jsmariegaard if you like I can add those things to the PR (not sure if I have push rights for mikeio though)

After I wrote this comment, @ecomodeller told me that pathlib Paths are supported with this implementation. But @PaulDanielML , you experienced something else, right?

PaulDanielML:
Yes, but of course yesterday I was using the mikeio version based on main, which raised an error when I passed a Path object. Looking at the implementation in this PR, it seems os.path.splitext can handle Path objects fine in order to get the extension 👍

PR #533, opened 2023-01-12
Dfsu 3D append fix

cmitR:
Dfsu 3D large files related update on the _write() and append() methods in dfsu.py to acount for 3D geometry


cmitR:
Right. Updated and was good it failed initially as the geometry was written wrong.

ecomodeller (approved):

Issue #539, opened 2023-01-21
Wrong location returned when accessing data in dfsu file using read


macrophanerophyte:
**Describe the bug**
When reading dfsu sigmaz files with ds = mikeio.open(file); col = ds.read(x=x, y=y), it happens that the position returned is rather far from the (x, y) point. 

**To Reproduce**
This happens when the the (x,y) point is close to a node and therefore is not contained in the 2 nearest elements to the point (step 3 in FM_Geometry._find_element_2d). I think in the function _find_element_containing_point_2d of "FM_Geometry", the node coordinates should be obtained via "self._geometry2d.node_coordinates" and not "self.node_coordinates". 

At line 731 of FM_Geometry: ids[k] = many_nearest[lid] if lid > 0 else -1
-1 is returned if no element are found to contain the point, which is then interpreted as the index of the last element and returns the wrong location

**System information:**
 - MIKE IO version: 1.2.4

jsmariegaard:
Thanks for reporting. We will need to fix this.

ecomodeller:
@macrophanerophyte  great username 🚀

ecomodeller:
Fixed in #541

PR #541, opened 2023-01-24
Fix find index (i.e. don't use -1)

ecomodeller:
#539


ecomodeller:
@jsmariegaard I changed the signature for `_find_element_2d(self, x: float, y: float) -> int:` to make it easier to reason about.

TL;DR we have not properly tested the behaviour for points outside the domain😳

jsmariegaard:
I see that you did a major revision of _find_element_2d(). It looked the way it did, after quite a lot of optimization work trying to make it as fast as possible for many points on large meshes. So maybe you could do a few benchmark tests? Or maybe we should do a c-implementation of this anyway...

ecomodeller:
Results from first "benchmark"
image

jsmariegaard:
Could you do the same for the main branch?

ecomodeller:
## Code qualities:
1. Correct
2. Predictable
3. Fast
image

ecomodeller:
`_find_nearest_2d_elements` is pretty fast thanks to https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html
image

And you know what Knuth thinks about premature optimization...

ecomodeller:
@jsmariegaard Can we merge this?

jsmariegaard:
I'll have a look today

mikeio/spatial/FM_geometry.py
```diff
@@ -804,47 +810,52 @@ def _find_element_2d(self, coords: np.array):
...
+                    nodes = self._geometry2d.element_table[p]
+                    element_found = self._point_in_polygon(
+                        nc[nodes, 0], nc[nodes, 1], coords[k, 0], coords[k, 1]
+                    )
+                    if element_found:
+                        ids[k] = p
+                        break
+
+                if not element_found:
+                    raise OutsideModelDomainError(x=coords[k, 0], y=coords[k, 1])
```
jsmariegaard, line 836: add k as info to error message. And even better - loop over all coords first and find out which are outside and provide user a list of all points outside so he/she does not need to call this method many times just to find out that the next point is also outside

mikeio/spatial/FM_geometry.py
```diff
@@ -1097,21 +1110,16 @@ def isel(self, idx=None, axis="elements", keepdims=False):
...
-            else:
-                return self.elements_to_geometry(elements=idx, node_layers=None)
+            return self.elements_to_geometry(
+                elements=idx, node_layers=None, keepdims=keepdims
+            )
 
-    def find_index(self, x=None, y=None, coords=None, area=None):
-        """Find element indicies for a number of points or within an area
+    def find_index(self, x=None, y=None, coords=None, area=None) -> Set[int]:
+        """Find a *set* of element indicies for a number of points or within an area
```
jsmariegaard, line 1122: Maybe state that this means that the returned collection is unordered and unique? (most people would know what a set is, but we also have many in-experienced python users).

mikeio/spatial/grid_geometry.py
```diff
@@ -644,18 +646,16 @@ def _xy_to_index(self, xy):
         y = xy[:, 1]
         x = xy[:, 0]
 
-        ii = (-1) * np.ones_like(
-            x, dtype=int
-        )  # TODO -1 is not a good choice, since it is a valid index
-        jj = (-1) * np.ones_like(
-            y, dtype=int
-        )  # TODO -1 is not a good choice, since it is a valid index
+        ii = (-999999999) * np.ones_like(x, dtype=int)  # TODO -999999999 ?
```
jsmariegaard, line 649: remove TODO

jsmariegaard (changes requested):
Lots of good stuff. But generally, I would prefer if find_index returns an np.array of int. If I give you an array of 100 positions (stations), I would like to have an array of indices back (similar to `contains()` ), not an unordered set. If I use the isel method I would get a DataArray/Dataset back for which a geometry would have been constructed and in that case the ordering should dissapear (like the non-uniqueness of found points). Let's talk tomorrow.

jsmariegaard (changes requested):
Still not happy about returning a set when you have a list of stations that you would like to find the corresponding indicies for - how do you recommend the user should get this now? 

I can see the problem about that find index takes both an area and a coords argument and that it is difficult to return the same datastructure in both cases... I would prefer to get an array with the same order as the input if possible and otherwise simply ordered by id. 

If you want to keep the current implementation (and I guess you will), then you should be consistent: right now only 3 out of the 6 find_index methods return a set:

image

ecomodeller:
> Still not happy about returning a set when you have a list of stations that you would like to find the corresponding indicies for - how do you recommend the user should get this now?
> 
> I can see the problem about that find index takes both an area and a coords argument and that it is difficult to return the same datastructure in both cases... I would prefer to get an array with the same order as the input if possible and otherwise simply ordered by id.
> 
> If you want to keep the current implementation (and I guess you will), then you should be consistent: right now only 3 out of the 6 find_index methods return a set:
> 
> image

We need two methods
* one method to return a indices of a subset which we can construct a meaningful geometry from, this will only work for FMGeometries and potentially for a connected subset of a gridded geometry (we can not support disconnected grid geometries)
* another method to return indices for every point in a list of points

And these methods have to be named in a meaningful way.

ecomodeller:
> > Still not happy about returning a set when you have a list of stations that you would like to find the corresponding indicies for - how do you recommend the user should get this now?
> > I can see the problem about that find index takes both an area and a coords argument and that it is difficult to return the same datastructure in both cases... I would prefer to get an array with the same order as the input if possible and otherwise simply ordered by id.
> > If you want to keep the current implementation (and I guess you will), then you should be consistent: right now only 3 out of the 6 find_index methods return a set:
> > image

`_GeometryFMLayered.find_index` and `GeometryFMVerticalProfile.find_index` uses [`np.intersect1d`](https://numpy.org/doc/stable/reference/generated/numpy.intersect1d.html) which returns "..the sorted, unique values...". Should we return this instead of a `set[int]` for all methods? 

My point is that unique values is not the same as a list of indices for each point, since many of the input locations will likely belong to the same element.

ecomodeller:
And while looking at spatial subselection in 3d files, there seems to be two methods with identical purpose, one used by the `Dfsu.read` and the other used by `Dataset.sel`
image

`_parse_geometry_sel` delegates to `find_index` but not with all arguments 🤔

PR #546, opened 2023-02-09
Make Dataarray/Dataset type consistent

ecomodeller:
```console
$ mypy mikeio/dataarray.py
Success: no issues found in 1 source file
```
These changes had some consequences, e.g. getting a subset with a boolean mask is not possible using `DataArray.__getitem__` since this normally returns a new `DataArray` and not a `np.ndarray`


ecomodeller:
The aggregation across items returns a DataArray which is inconsistent with other types of aggregations, but make sense if the aggregation is over all items. In a future version I will implement another argument to inform which items->item to do the aggregation over. The input file might have ensemble members for water level, and currents and they have to be aggregated into separate items.

mikeio/dataarray.py
```diff
@@ -483,10 +484,14 @@ def squeeze(self) -> "DataArray":
 
     # ============= Select/interp ===========
 
-    def __getitem__(self, key) -> "DataArray":
+    def get_masked(self, key) -> np.ndarray:
```
ecomodeller, line 487: @jsmariegaard Since I don't want `__getitem__` to return anything else than a DataArray. This is the solution so far...

Can you remind me of the use case for filtering with a boolean mask?

Is this the way to go or should we keep it in `__getitem__`  and return the result as a DataArray instead (with proper time and geometry)?

jsmariegaard: Often you would want to set a value based on a criteria, but this is done with __setitem__, so not affected by this change. The use case with __getitem__ could be that you want to do some statistics based on a criteria, e.g. "for the waves coming from North, what is the mean wave height"... It is not possible to return a DataArray directly with these values only, as the geometry could be different for each timestep. One option could be to return a DataArray where all the values not meeting the criteria have been replaced with NaN - this is similar to xarray where method https://docs.xarray.dev/en/stable/generated/xarray.Dataset.where.html but here you have the option to set the "other" value. Maybe we should create a where method instead of get_masked???

ecomodeller: The xarray Dataset.where returns the same data type as self, https://github.com/pydata/xarray/blob/main/xarray/core/common.py#L996-L1019, if you call it from a DataArray, you get a DataArray back, if you call it from a Dataset, you get a Dataset back.

PR #547, opened 2023-02-13
Read items by name fix

Hendrik1987:
I get the to me unexpected behavior:
After reading items by name
```
items = ["NotFun", "VarFun01"]
dfs = Dfs0(dfs0file)
dfs.read(items=items) 
```
`items` is no longer equal `["NotFun", "VarFun01"]`


ecomodeller:
Nice spotted, mutating inputs is generally a bad idea, since it is so unexpected.

PR #551, opened 2023-03-07
Private modules

ecomodeller:
In response to https://github.com/DHI/mikeio/pull/523 on what is private and what is public, This PR will make it clearer which classes and functions that are supposed to be part of the public API.

```python
# This class is for internal use only indicated by the leading underscore of the _grid_geometry module
mikeio.spatial._grid_geometry.BoundingBox

# While this is the public API
mikeio.Grid2D
```


ecomodeller:
File/module hierarchy inspired by e.g. https://github.com/scikit-learn/scikit-learn/tree/main/sklearn/

Issue #559, opened 2023-03-27
read_pfs fails in [WQRiverGroupPfs] section for .mhydro file


Snowthe:
I am trying to read a .mhydro file (MIKE Zero 2022 SP1) with mikeio.read_pfs(), which fails with the following error:
```
ValueError: DK1_100m_2023_Nsj_Lakes_Trial.mhydro could not be parsed. while scanning a block scalar
  in "<unicode string>", line 72225, column 80:
,0,"",0,"",0,"",0,"",0,"",0,"",'||',false>'
                                ^
expected chomping or indentation indicators, but found '|'
  in "<unicode string>", line 72225, column 81:
0,"",0,"",0,"",0,"",0,"",0,"",'||',false>'
                                ^
```
It fails when reading the [WQRiverGroupPfs] section (which in my .mhydro setup should just be inactive/unused). The whole section in my .mhydro file looks like this:

```
   [WQRiverGroupPfs]
      Touched = false
      [WQRiverListPfs]
         Touched = false
         DhiSEPfsListItemCount = 1
         SolutionMethod = 1
         UpdateFrequency = 1
         DisableProcesses = false
         [WQRiverPfs_0]
            Touched = false
            ObjectID = '7669105c-471a-4b76-be8b-4978926ffe10'
            ModelIndex = 0
            Model = 'From file ...'
            TemplateFile = ||
            Description = ''
            Clob = '<CLOB:22,1,1,false,1,0,"",0,"",0,"",0,"",0,"",0,"",0,"",0,"",||,false>'
         EndSect  // WQRiverPfs_0

      EndSect  // WQRiverListPfs

   EndSect  // WQRiverGroupPfs
```

...and it actually fails in the line " Clob = '<CLOB:22,1,1,false,1,0,"",0,"",0,"",0,"",0,"",0,"",0,"",0,"",||,false>' ", referring to the pipe symbols.

**System information:**
 - Python 3.8.11
 - MIKE IO 1.4.1

PR #560, opened 2023-03-27
Support CLOB in `read_pfs()`

ecomodeller:
#559


Snowthe:
> @Snowthe is this a suitable solution?

Yes, thanks a lot for the quick fix - as far as I can tell, this is working

jsmariegaard (approved):
Great idea with the "context". Well done!

PR #573, opened 2023-06-01
add _{item_#} for repeated item names

mohm-dhi:
currently when the item name are the same it throws an error, instead I change the code so it modify the item name by adding _2, _3, _4 etc to the repeated item name.


daniel-caichac-DHI:
I have had a similar issue before mainly due to mike's limitation on the number of characters (if your Dfs0 point output name is too long, I think it truncates to 32 characters or so, and I end up with repeated name items in my Dfs0). This does not allow you to read the items as mikeio is right now.
What I've done is read only a few items (say items 1-15), then rename in a dataframe, and then read the remaining items (say items 15-20 which were the ones with repeated names).
I agree with Henrik that allowing identical names is fragile, but sometimes necessary.

ecomodeller:
Added warning:
```
>>> import mikeio
>>> ds = mikeio.read("tests/testdata/untitled_3_items.dfs0")
/home/jan/src/mikeio/mikeio/dataset/_dataset.py:179: UserWarning: Duplicate item name: Untitled. Renaming to Untitled_2
  warnings.warn(f"Duplicate item name: {item}. Renaming to {item}_{count_dict[item]}")
/home/jan/src/mikeio/mikeio/dataset/_dataset.py:179: UserWarning: Duplicate item name: Untitled. Renaming to Untitled_3
  warnings.warn(f"Duplicate item name: {item}. Renaming to {item}_{count_dict[item]}")
```

PR #592, opened 2023-08-10
Type hints from the future

ecomodeller:
This line enables the union type hint syntax from 3.10 to be used in 3.7+
```python
from __future__ import annotations
```

I.e.

We can replace
```python
def foo(Union[int,str]) -> Union[int,bool):
  ...
```
with the version that is easier to read for humans🙂
```python
def foo(int | str) -> int | bool
  ...
```

### DIFF
--- a/mikeio/_track.py
+++ b/mikeio/_track.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
 import os
 from datetime import datetime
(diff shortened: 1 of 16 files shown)


ecomodeller:
And remove some deprecated things, mainly the aggregation on Dataset, which now consistently returns a new Dataset, which should be helpful for auto-completion.

PR #593, opened 2023-08-11
New notebook with conversion from Esri ASCII raster to dfs2

Q-R-B:
Adding a notebook on how to convert Ascii files to Dfs2.


ecomodeller:
⭐ Great with some real-world example! 🚀 

Some reflections:

* Consider a more descriptive title of the PR
* Is this data publicly available (even if it is in a different format)? Can we cite the source?
* In the notebook you print the full contents of the ascii file, but I don't think it is necessary in order to understand the process.
  Isn't it enough to print the header and a few lines of the data in order to understand the data format?
* Mention that the format is a specific Ascii format, [Esri ASCII raster format](https://desktop.arcgis.com/en/arcmap/latest/manage-data/raster-and-images/esri-ascii-raster-format.htm)?

Q-R-B:
thanks for the comments, 

No data is not publicly available so i'll find something more appropriate.
and i'll fix the rest as well.

What is the best way to do this? 
Should i close the pull request and create a new one?

ecomodeller:
There is no need to close the PR, just push changes to this branch and once we agree that is complete, I will merge the changes.

Q-R-B:
Ok, i think it's good now, 
there is just a typo in the commit "Adding am Ascii file" that should be "Adding an Ascii file". I don't seem to be able to modify it.

ecomodeller (approved):

PR #597, opened 2023-09-08
Corrected contour generation

Havrevoll:
The for loop did only read one contour at each level, but often, (also in the example shown), there are several contours on each level, so the program must iterate over all contours and create a dictionary of polygons for each wind speed.


ecomodeller:
@Havrevoll when changes are complete, (maybe already) just ask for a review 
image

ecomodeller (approved):
Great.
With this change we combine two arrays and looping over two arrays at the same time can be done with the `zip()` function.

E.g.:
```python
for collection, level in zip(cn.collections, contour_levels):
    for p in collection.get_paths():
        v = p.vertices
        x = v[:,0]
        y = v[:,1]
        poly = LineString([(i[0], i[1]) for i in zip(x,y)])
        if(poly.is_empty):
            print(f"{i} is empty")
        poly_list.append(dict(wind_speed = level, poly=poly))
```

I also realized that the GeoDataframe is missing the coordinate reference system.
image

PR #608, opened 2023-10-12
Overload `Dataset.__getitem__`

ecomodeller:
In order to make better use of type hints for auto-completion.

Selecting a single item returns a `DataArray`, while selecting multiple items returns a `Dataset`.

Consistent with `xarray`.


ecomodeller:
image

jsmariegaard (approved):
Cool!

PR #611, opened 2023-10-18
Pfs improvements

jsmariegaard:
Minor tweaks and fixes to PFS module


mikeio/pfs/_pfsdocument.py
```diff
@@ -295,9 +301,14 @@ def _parse_line(self, line: str, level: int = 0) -> Tuple[str, int]:
                 value = s[(idx + 1) :].strip()
 
                 if key == "start_time":
-                    value = datetime.strptime(value, "%Y, %m, %d, %H, %M, %S").strftime(
-                        "%Y-%m-%d %H:%M:%S"
-                    )
+                    try:
+                        value = datetime.strptime(value, "%Y, %m, %d, %H, %M, %S").strftime(
```
ecomodeller, line 305: I think we should re-evaluate if MIKE IO should attempt to parse datetimes from PFS files.

It seems like an incomplete solution to only parse keys named `start_time` (used by the FM engine) and formatted as a six comma separated integers.

When there are more datetimes and formats in pfs files, e.g.

https://github.com/DHI/mikeio/blob/dab4fb9b9e626418a15c2c289dadfdc7017787a4/tests/testdata/pfs/Karup_basic.she#L263

https://github.com/DHI/mikeio/blob/dab4fb9b9e626418a15c2c289dadfdc7017787a4/tests/testdata/pfs/Setup_WaveSpectraConverter.21t#L8C5

PR #613, opened 2023-10-26
More static types

ecomodeller:
Be more strict...


ecomodeller (commented):
Use | None instead of Optional consistently

PR #619, opened 2023-11-07
Accept dfs file with zero timestep

ecomodeller:
Some engines, e.g. MIKE She specifies a file $\Delta t = 0.0$, which could not be correctly interpreted by MIKE IO (actually `pd.DateTimeIndex`).

This PR warns when $\Delta t <= 0.0$ or raises an error if number of timesteps > 1 (where the files would be invalid).

Example of reading a previously unreadable file:
image


ecomodeller:
Bad starting point😶

PR #622, opened 2023-11-08
Simplify Dfs{0,1}

ecomodeller:
Removes syntax of creating an instance of `Dfs0()` without a filename and then write data to it.


jsmariegaard:
Good idea 😃

PR #632, opened 2024-01-11
Isel should respect order

ecomodeller:
For some advanced use cases it can be beneficial to extract a subset of elements in a particular order. #631


ecomodeller:
Superseded by #690

PR #650, opened 2024-02-14
Create better xarray from common MIKE Dfs0 output

ecomodeller:
image


ecomodeller:
@daniel-caichac-DHI do you think this feature would be useful?

PR #660, opened 2024-03-01
Experiment: Dfsu - vertical aggregation

ecomodeller:


ecomodeller:
@jsmariegaard what do you think about the idea of reshaping Dfsu 3d data into (time, layers, element_2d)?

For sigma only models, this shape seems very natural, for sigma-z, less so, as it doesn't always fit perfectly into a NumPy cube. 🤔

jsmariegaard:
For sigma-only it would be super nice, but how would you handle the sigma-z situation? Often you would have many-many layers (100s) in a small part of the domain... 

Previously I was thinking we could have a virtual 3d shape - so that you could index like it was 3d but we keep the shape as is?

PR #662, opened 2024-03-03
Fix PfsDocument.copy

ecomodeller:
Inheritance is tricky...


ecomodeller:
😞 No good if we don't have write access to temp folder🤔

PR #663, opened 2024-03-04
More type hints

ecomodeller:


ecomodeller:
* Quite challenging to add static type information to functions with multiple return types.
* A bit challenging to add proper types related to NumPy (MIKE IO relies very heavily on NumPy).

PR #682, opened 2024-03-25
Static item bathymetry

ecomodeller:


ecomodeller:
Based on my first naive implementation I discovered that MIKE SHE also has some static items.

Any comments on the usefulness of those from @watermain or @Snowthe?

watermain:
I´m not sure many people are aware these static items exist in mshe result files. They have some potentially useful information though.

Snowthe:
Sorry, I'm not following right now - what would the static items be compared to "normal" items in a MIKE SHE dfs2/dfs3 output?

I. however, now receive a new warning message when reading e.g. a PreProcessed file (i.e. one without timesteps):
  C:\miniconda3\envs\py38\lib\site-packages\mikeio\dfs\_dfs.py:155: UserWarning: Time step is 0.0 seconds. This must be a positive number. Setting to 1 second.
    warnings.warn(
I do not receive the same message when reading a result file with timesteps. Is that related to your latest updates (and this issue)?

watermain:
MShe just doesn´t bother to set a time step length > 0 for non-dynamic files/items. Mikeio is unhappy about this, but it just carries on. It used to be considered an error but that has been changed a while ago.

Possible static items from all output files are:
- X-location
- Y-location
- Z-location
- Topography
- Mapping from 1d cell index to x cell index
- Mapping from 1d cell index to y cell index
- Mapping from 1d cell index to UZ 1d cell index
- Mapping from UZ 1d cell index to 1d cell index
- Thickness of UZ compartment
- 1d index of the bottom UZ compartment of each column
- Hor. hydr. conductiivity in x-direction
- Hor. hydr. conductiivity in y-direction
- Vert. hydr. conductiivity
- spec. yield
- storage coefficient
- index of SZ boundary cell
- bathymetry (mimicking an m21  output file)

Snowthe:
> MShe just doesn´t bother to set a time step length > 0 for non-dynamic files/items. Mikeio is unhappy about this, but it just carries on. It used to be considered an error but that has been changed a while ago.

Yes; however the warning message is new (v2.0 beta; maybe already with v1.7?). Whereas before (v1.6.2), I simply do not get any error or warning; it just loads.

> Possible static items from all output files are: [...]

So to understand - those items are all stored with a MIKE SHE dfs3 result file (e.g. the 3D_SZ.dfs3 file)? That information can also be accessed from the PreProcessed file; however it definitely would be useful to have access to it in the same file.

watermain:
None of the output files has all of these static items, it´s usually only a few per file (dfs2 or dfs3), or maybe none in some cases. Also I cannot easily check if all these items are in at least one file, but these are the ones the mshe dfs layer knows about. Some of these could also be taken from the PP´ed files, others can´t.

Issue #685, opened 2024-04-10
Point incorrectly considered outside domain


coline-mathias:
**Describe the bug**
Hello, I just encountered a bug in mikeio-1.7.1 when trying to extract data in specific points. All the points work fine except one of them, for which data.sel() raises OutsideDomainError. I checked on a plot, and this point is in the middle of the domain, not especially close to a boundary or anything. 

**To Reproduce**
Below is an example with two points : for the first one the extraction is successful, the second one raises OutsideDomaineError. However, mesh.contains(pt) returns True.

`pts = [(439166.047,  6921703.975), (439297.166,  6921728.645)]`
`output_file= "SW.dfsu"`
`data = mikeio.read(output_file, items=['Sign. Wave Height', 'Mean Wave Direction', 'Peak Wave Period', 'Surface elevation'])`
`mesh = mikeio.Mesh(output_file)`
`for pt in pts : `
`      print(mesh.contains(pt))`
`      data_pt = data.sel(x=pt[0], y=pt[1])`

Link to SW data file : https://creocean.fromsmash.com/vZre2O85U0-ct

**Expected behavior**
I would have expected both points to be corrrectly extracted from data file. 

**Screenshots**
Position of the two points tested above :
![image](https://github.com/DHI/mikeio/assets/163521379/6aba692c-8253-4b8b-804b-7ec287e1df11)

**System information:**
 - Python version : '3.11.5 | packaged by Anaconda, Inc. | (main, Sep 11 2023, 13:26:23) [MSC v.1916 64 bit (AMD64)]'
 - MIKE IO version: 1.7.1

Thank you very much,

Coline

ecomodeller:
Our current strategy to find the correct element (trying to be as fast as possible) is not correct in areas with such a mixed element sizes.

The failing point is not inside the any of the 10 closest elements and then wrongly categorized as outside the domain.
![image](https://github.com/DHI/mikeio/assets/614215/80be4e47-ffc4-4cbd-bd35-1573a1edffd6)

https://github.com/DHI/mikeio/blob/08f52ac9fb6f2c91599f56a5f62a277b296c7dde/mikeio/spatial/_FM_geometry.py#L688
You can tell from the comment, that I was a bit unsure on how to handle this, but this example indicates that 10 is not enough😳

ecomodeller:
Attaching a smaller file of the AOI to make it easer to reproduce.
[subset.zip](https://github.com/DHI/mikeio/files/14931018/subset.zip)

PR #700, opened 2024-06-05
Fix GeometryFM.contains and expand use of shapely

ecomodeller:


jsmariegaard:
I guess it is not possible to use _point_in_polygon() ?

ecomodeller:
> I guess it is not possible to use _point_in_polygon() ?

I suppose we can, but that is only part of the solution, we still need to construct a proper polygon with holes for the entire domain, unless we loop over all elements, which we could do in a clever way similar to how `find_index` works (when #685 has been fixed) .

ecomodeller:
@jsmariegaard 
[...plot.outline()](https://dhi.github.io/mikeio/api/spatial._FM_geometry._GeometryFMPlotter.html#mikeio.spatial._FM_geometry._GeometryFMPlotter.outline) is used to plot the outline of the model domain. Very useful!
image


We have the ability to extract elements into a new file, e.g.
image

But the subset has lost the reference to the original model domain.

Does the concept of `outline` of the model domain still apply for a subset?

jsmariegaard:
> But the subset has lost the reference to the original model domain.
> 
> Does the concept of `outline` of the model domain still apply for a subset?

I guess it does *not* in many cases. But the user would probably know that and could accept that the plot.outline() would be empty or raise an error 🤔

PR #702, opened 2024-06-12
Non-equidistant time axis for Dfsu

ecomodeller:


FrejaTerpPetersen:
Sorry for a very slow reply to this PR.
As far as I can see (and as I tested), I can now create a Dataset with a non-equidistant time axis. I was also able to concatenate two datasets resulting in a non-equidistant time axis. The same goes for dataarrays.  This is very useful for my use of mikeio, so I will be very happy with this update. :) I didn't meet any issues.

PR #714, opened 2024-08-15
Fixing copying of data_value_type in from_pandas

srwater:
also changing tests to cover this change and fixing doc typo in _dataset.py


ecomodeller:
Thanks @srwater for this contribution!

PR #719, opened 2024-08-21
Special case for generic.concat dfs0

ecomodeller:


daniel-caichac-DHI:
no unit test to check that it actually works?

daniel-caichac-DHI (approved):
Looks good

daniel-caichac-DHI:
Someone else needs to approve though, I just did but I do not have privileges to approve apparently

PR #730, opened 2024-10-08
Slim sdist

ecomodeller:
Only keep the necessary files in the source distribution., (Wheel was already minimal).


ryan-kipawa (approved):
⛴️

PR #732, opened 2024-10-15
Grid docstrings

ecomodeller:


jsmariegaard:
Thanks - I am still a bit confused about this (issue + PR). Why did you remove the axis_name? Is it no longer possible to indicate that an axis is "z" instead of "x" as an example? And x0? 

Now I see you moved the docstring from the top of class to the init method, but will I then get inline help, when I am creating a new Grid1D? (and I agree that I did confuse the arguments of the initialiser with the properties of the class)

ecomodeller:
`_axis_name` and `_x0` are private attributes, do you think they should be read-only properties? We could make `axis_name` a settable property, if it is useful?

```python
>>> from mikeio import Grid1D
>>> g = Grid1D(x=[0,1], axis_name="z")
>>> g
<mikeio.Grid1D>
x: [0, 1] (nx=2, dx=1)
>>> g._axis_name
'z'
>>> g._x0
np.int64(0)
>>> g2 = Grid1D(x0=0.1, dx=1, nx =2, axis_name="y")
>>> g2
<mikeio.Grid1D>
x: [0.1, 1.1] (nx=2, dx=1)
>>> g2._x0
0.1
>>> g2._axis_name
'y'

jsmariegaard (approved):

PR #739, opened 2024-11-01
generic_average

daniel-caichac-DHI:
I got it semi working, but not yet.
I need help with mikeCore I think.

### DIFF
--- a/mikeio/generic.py
+++ b/mikeio/generic.py
@@ -526,8 +526,8 @@ def concat(
                         darray = d.astype(np.float32)
 
                         dfs_o.WriteItemTimeStepNext(0, darray)
-                end_time = (
-                    start_time + timedelta(seconds=timestep * dt)
+                end_time = start_time + timedelta(
+                    seconds=timestep * dt
                 )  # reuse last timestep since there is no EndDateTime attribute in t_axis.
                 dfs_i.Close()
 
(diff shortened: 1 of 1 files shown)


mikeio/generic.py
```diff
@@ -552,6 +552,79 @@ def concat(
...
+                                    temp_dfs = DfsFileFactory.DfsGenericOpen(
+                                        str(outfilename)
+                                    )
+                                    existing_data = temp_dfs.ReadItemTimeStep(
+                                        int(item + 1), int(existing_pos)
+                                    ).Data
+                                    temp_dfs.Close()
+
+                                    # Calculate average
+                                    averaged_data = (existing_data + current_data) / 2
```
daniel-caichac-DHI, line 605: This is calculating the average as soon as the dfs2's overlap

mikeio/generic.py
```diff
@@ -552,6 +552,79 @@ def concat(
...
+
+                                    # Write averaged data
+                                    dfs_o.WriteItemTimeStepNext(0, averaged_data)
```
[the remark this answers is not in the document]
ecomodeller: @ryan-kipawa Is this something that you would like to dig into?

mikeio/generic.py
```diff
@@ -552,6 +552,79 @@ def concat(
...
+                                    (current_time - first_start_time).total_seconds()
+                                    / dt
+                                    - 1
+                                )
+                                if existing_pos >= 0:
+                                    # Read existing data
+                                    dfs_o.Flush()
+                                    temp_dfs = DfsFileFactory.DfsGenericOpen(
+                                        str(outfilename)
+                                    )
```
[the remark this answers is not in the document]
daniel-caichac-DHI: image

mikeio/generic.py
```diff
@@ -552,6 +552,82 @@ def concat(
                 )  # get end time from current file
                 dfs_i.Close()
 
+        if keep == "average":
+            if i == 0:
+                # For first file, write all timesteps normally
```
ecomodeller, line 557: Isn't it better to calculate the timestep where the overlap occurs, calcuate the average, write that and then the rest, instead of writing the entire contents and then trying to overwrite?

jsmariegaard: Agree. I would suggest an algorithm like this (file1 starts before file2): 

* open file1
* write file1 timesteps until start of file file2
* open file2
* write overlapping steps by with alpha*data1 + (1-alpha)*data2 until end of file1
* close file1
* write rest of file2 
* close file2 

using the alpha notation above it will be super simple to later add a keep="linear" that linear mixes the two in the overlapping period.

daniel-caichac-DHI: don't shoot the messenger, I chat'gpted this solution due to my lack of mikecore competences

ecomodeller:
@daniel-caichac-DHI is this PR still relevant?

daniel-caichac-DHI:
I never managed to actually make it work, and with the amount of conflicts, it might be quicker to start from scratch.
I will close it

Issue #749, opened 2024-11-15
Very slow reading of mhydro file


Havrevoll:
**Describe the bug**
I have three different mhydro files, of sizes 2-5 MB. When reading them with `mikeio.read_pfs`, one takes 2 minutes to read, one takes 10 minutes and the last takes 20 minutes to read, and I can't understand why it takes so long. Are there any catchment polygons or other data that are unusually complicated to parse? After the files are read into memory, all reading and changing of values goes quickly. Writing the file again is done in no time.

**To Reproduce**
Load the relevant file, sent by email to JEM and JAN:
`mhr_file = mikeio.read_pfs("river_name.mhydro")`
I don't think it matters if I use IPython, Jupyter Notebook or plain Python.

**Expected behavior**
The file would be loaded into memory in some seconds.

**System information:**
 - Python version 3.12.5
 - MIKE IO version 1.6.3

ecomodeller:
Initial profiling has revealed that ~100% of the time is spent in 
https://github.com/DHI/mikeio/blob/299e509cf71b7b2193c2f0a99531ffcf7053e034/mikeio/pfs/_pfsdocument.py#L361-L364

@jsmariegaard the name of the private method `_split_line_by_comma` suggests that it should be simple, but the reason the regex is needed is to avoid splitting on commas inside strings.

There are some really long lines in the example file, one line contains a line like this:
```
Shape = 'MULTIPOLYGON(((475130.002457948 6524999.99673699,4751
```
with **25799** commas inside it, I don't know if these are the ones that takes time, but at least it contains many commas.

jsmariegaard:
I wonder if we could have special handling of a `MULTIPOLYGON` 🤔?

ecomodeller:
> I wonder if we could have special handling of a `MULTIPOLYGON` 🤔?

To try this out I removed the 11 lines with MULTIPOLYGON
```bash
grep -v MULTIPOLYGON file_name.mhydro
```

Parsing the original pfs file took ~10 minutes, the stripped file with 8992 lines (11 lines shorter) took 0.3 seconds to read🤯.

ecomodeller:
Special handling of `MULTIPOLYGON' brings the time down to 1.0 s
![image](https://github.com/user-attachments/assets/467b0a96-e487-4e14-b4cf-4c6fda383f73)

PR #750, opened 2024-11-19
Pfs - handle slow reading of MIKE Hydro

ecomodeller:
MIKE Hydro pfs files have catchment defined as WKT `MULTIPOLYGON` which can have many commas.

Closing #749


jsmariegaard:
Nice simple solution

ryan-kipawa (approved):
A bit verbose, but looks good 😛

PR #753, opened 2024-11-20
Added download button

daniel-caichac-DHI:
on readme.md

Quite straight forward


jsmariegaard:
I think @ecomodeller should take the decision on this one.

ecomodeller:
It is a small badge next to the other badges, but to whom does the information add value?

The metric can only be seen in relation to other packages, the actual number is not that useful since most downloads are from automated processes like CI and docker builds.

ecomodeller:
Numpy has download stats on their readme.

ecomodeller (approved):

daniel-caichac-DHI:
Well I did not have a chance to answer but yeah numpy and others.
I think it can give you a relative number of how many people/bots are downloading it, also compared to another libraries (modelskill, watobs) and maybe hint towards where to add more resources/efforts, even if the numbers are pumped by some bots.
Just saying

PR #761, opened 2024-11-28
Bugfix: bug fix in spectral 2D grid y ax

stkistner:
**Problem**: reading spectral Dfs2 with geometric projection CRS in header incorrectly applies the Easting origin to the directional axis.
**Reason**: missing if statement in `Grid2D.y`


ecomodeller:
Thanks @stkistner.

Grateful for the fix, but even better would be:
1. A failing test 🔴
2. Fix the code 🐍
3. Successful test ✔️
4. Refactor (not relevant in this case) 🧑‍💻

Otherwise it is hard to verify that the changes actually solves the reported bug.

stkistner:
@ecomodeller added file and passing test🙂 (will fail without fix)

PR #778, opened 2025-01-20
Fix keepdims

ecomodeller:


mikeio/__init__.py
```diff
@@ -121,11 +121,10 @@ def read(
     ext = Path(filename).suffix.lower()
 
     if "dfs" not in ext:
-        raise ValueError("mikeio.read() is only supported for Dfs files")
+        raise ValueError("mikeio.read() is only supported for dfs files")
 
     dfs = open(filename)
-    if isinstance(dfs, Mesh):
-        raise ValueError("mikeio.read() is not supported for Mesh files")
+    assert not isinstance(dfs, Mesh)
```
jsmariegaard, line 127: I do not understand this change - isn't the user better supported by a ValueError with a clear error message?

mikeio/dfs/_dfs.py
```diff
@@ -391,10 +391,16 @@ def read(
         elif self._ndim == 2:
             shape = (nt, self.ny, self.nx)  # type: ignore
         else:
+            # TODO this is not used, since Dfs3 has a separate .read method
             shape = (nt, self.nz, self.ny, self.nx)  # type: ignore
 
+        spdims = self.geometry.default_dims
+        dims = ["time"]
```
jsmariegaard, line 398: It seems a little confusing to me to first always prepend dims with "time" and then afterwards potentially removing it in line 403 🤔 But it looks like it works and you have probably considered alternative ways.

mikeio/dfs/_dfs.py
```diff
@@ -406,16 +412,19 @@ def read(
...
+                assert d.ndim == 1
 
                 d[d == self.deletevalue] = np.nan
 
                 if self._ndim == 2:
                     d = d.reshape(self.ny, self.nx)  # type: ignore
 
                 if single_time_selected:
-                    data_list[item] = d
+                    if keepdims:
```
jsmariegaard, line 424: Why not 

    data_list[item] = np.atleast_2d(d) if keepdims else d

tests/test_consistency.py
```diff
@@ -574,3 +574,38 @@ def test_concat_dfsu3d_single_timesteps_generic_vs_dataset(tmp_path):
     assert ds3.end_time == ds4.end_time
     assert ds3.n_items == ds4.n_items
     assert ds3.n_timesteps == ds4.n_timesteps
+
+
+def test_keepdims_removes_singleton_dimension() -> None:
```
jsmariegaard, line 579: Great tests ❤️

And I guess it is not relevant to add one with dfs0 or dfs3? (what happens with the keepdims argument then)

jsmariegaard (approved):
Great 

See comments and decide if you want to change :-)

mikeio/__init__.py
ecomodeller: True

mikeio/dfs/_dfs.py
ecomodeller: There is room for improvement...

tests/test_consistency.py
ecomodeller: I suppose dfs3 could be relevant. Let's revisit it when we improve dfs2 functionality.

Issue #482, opened 2022-11-23
Smart extraction / interpolation of points from large dfsu


Hendrik1987:
I still struggle to extract data from larger-then-memory .dfsu files.

Or maybe I am just not aware of the newest way to do this in mikeio?

**What I have tried / done so far:**
1. read only first timestep of .dfsu
2. identify n nearest elements
3. read all data only for the identified elements
```
dfs = mikeio.open(fn)
mds0 = dfs.read(time=0)
elements = mds0.geometry.find_nearest_elements(x=pts.x, y=pts.y, n_nearest=3)
elements_load = np.unique(elements)
mds = dfs.read(elements=elements_load)
mds[0].plot()
```
![image](https://user-images.githubusercontent.com/29595985/203555320-9bf80696-b619-40ea-84e8-a88552253739.png)

5. loop over the query points and interpolate from n nearest elements:
```
mdsi =  [mds.interp(x=x, y=y, n_nearest=3) for x, y in zip(pts.x, pts.y)]
```
![image](https://user-images.githubusercontent.com/29595985/203555525-1cea3e15-404a-43e0-b283-34587e419bcd.png)

**What I think could be enhanced:**
1. Make this (or a better) workflow easily available, maybe via `mikeio.read(x, y, n_nearest = 3)` 
2. Is there a way to then concat the `mikeio.Datasets` in the list `mdsi`, maybe resulting in a `GeometryPoint2DCollection` geometry?
3. How do I write the extracted points to `.dfs`, while maintain geometry information?

**Possibly related to fmskill enhancement**
- Here I also struggle to extract from large `.dfsu` model results. Maybe some of the same logic applies here?
- ModelResult should be able to mikeio.Dataset/DataArray (https://github.com/DHI/fmskill/discussions/132)

ecomodeller:
[...]

PR #779, opened 2025-01-21
Add missing dfsu read x,y test

ecomodeller:
`Dfsu2DH.read(x=, y=)` can be used to read a number of positions (stations).

Inspired by discussion in #482 

image

@Hendrik1987 is this good enough, or would you still prefer a `extract_stations` method?

### DIFF
--- a/mikeio/dfsu/_dfsu.py
+++ b/mikeio/dfsu/_dfsu.py
@@ -421,7 +421,7 @@ class Dfsu2DH:
             Read only data inside (horizontal) area given as a
             bounding box (tuple with left, lower, right, upper)
             or as list of coordinates for a polygon, by default None
-        x, y: float, optional
+        x, y: float or list[float], optional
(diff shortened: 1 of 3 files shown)


Hendrik1987:
@ecomodeller :
Thanks for keeping our old discussions in mind :-)

If your solution under the hood avoids reading all elements or in some other way handles larger than memory files, then this is amazing! I think, no need to introduce a dedicated `extract_station` method.

I rather think it could be useful to also here include a `n_nearest` argument, allowing a following interpolation methods on the `ds`. This would result in loading also these in addition to the ones in your plot above. In fact, the current ds interpolation does not define a max distance, or does it? I.e., blindly interpolationg on the subset you show above with default `n_nearest`, would lead to unexpected restults.

Hendrik1987:
@irnigerandrea: Don't you think this could be useful for us?

PR #789, opened 2025-02-06
{Dataset, DataArray}.fillna

ecomodeller:


jsmariegaard (approved):
Love it - only needs small section in docs IMO

ecomodeller:
> Love it - only needs small section in docs IMO

Above `dropna()` in this section https://dhi.github.io/mikeio/user-guide/dataset.html#mathematical-operations ?

PR #792, opened 2025-02-10
Support multiplication and division of datasets.

ecomodeller:
* Allow multiplication + division
* Be more permissive, only force shapes of the two datasets to be identical
* Use item info from the first argument (no longer invent new name e.g. "foo * bar")

### DIFF
--- a/mikeio/dataset/_dataset.py
+++ b/mikeio/dataset/_dataset.py
@@ -1341,20 +1341,6 @@ class Dataset:
                 f"Number of items must match ({self.n_items} and {other.n_items})"
             )
 
-        for this, that in zip(self.items, other.items):
-            if this.name != that.name:
-                raise ValueError(
-                    f"Item names must match. Item: {this.name} != {that.name}"
-                )
-            if this.type != that.type:
-                raise ValueError(
-                    f"Item types must match. Item: {this.type} != {that.type}"
-                )
-            if this.unit != that.unit:
-                raise ValueError(
-                    f"Item units must match. Item: {this.unit} != {that.unit}"
-                )
-
     # ============ aggregate =============
 
     def aggregate(
@@ -1779,16 +1765,10 @@ class Dataset:
     def _add_dataset(self, other: "Dataset", sign: float = 1.0) -> "Dataset":
         self._check_datasets_match(other)
         try:
-            data = [
-                self[x].to_numpy() + sign * other[y].to_numpy()
-                for x, y in zip(self.items, other.items)
-            ]
+            data = [x + sign * y for x, y in zip(self, other)]
         except TypeError:
             raise TypeError("Could not add data in Dataset")
-        newds = self.copy()
-        for new, old in zip(newds, data):
-            new.values = old
-        return newds
+        return Dataset(data)
 
     def _check_datasets_match(self, other: "Dataset") -> None:
         self._check_all_items_match(other)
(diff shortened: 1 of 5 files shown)


tests/test_dataset.py
```diff
@@ -1201,6 +1201,34 @@ def test_sub_dataset(ds1, ds2):
...
+    dsa = mikeio.Dataset({"Foo": mikeio.DataArray([1, 2, 3])})
+    dsb = mikeio.Dataset({"Foo": mikeio.DataArray([4, 5, 6])})
+    dsr = dsa * dsb
+    assert np.all(dsr[0].to_numpy() == np.array([4, 10, 18]))
+
+
+def test_multiply_datasets_must_match():
+    dsa = mikeio.Dataset({"Foo": mikeio.DataArray([1, 2, 3])})
+    dsb = mikeio.Dataset({"Bar": mikeio.DataArray([4, 5, 6])})
+    with pytest.raises(ValueError):
```
jsmariegaard, line 1214: Why do the name/item need to match on multiplication/division?

mikeio/dataset/_dataset.py
```diff
@@ -1830,6 +1837,29 @@ def _multiply_value(self, value: float) -> "Dataset":
...
+                self._check_datasets_match(value)
+                data = [
+                    self[x].to_numpy() / value[y].to_numpy()
+                    for x, y in zip(self.items, value.items)
+                ]
+            else:
+                data = [self[x].to_numpy() / value for x in self.items]
+        except TypeError:
+            raise TypeError(f"{value} could not be divided to Dataset")
+        items = deepcopy(self.items)
```
jsmariegaard, line 1852: I guess the unit is often not the same after division 🤔

tests/test_dataset.py
```diff
@@ -1201,6 +1201,34 @@ def test_sub_dataset(ds1, ds2):
...
+    ds_nom = mikeio.Dataset({"Foo": mikeio.DataArray([1, 2, 3])})
+    ds_denom = mikeio.Dataset({"Foo": mikeio.DataArray([4, 5, 6])})
+    ds3 = ds_nom / ds_denom
+    assert np.all(ds3[0].to_numpy() == np.array([0.25, 0.4, 0.5]))
+
+
+def test_divide_dataset_must_match():
+    dsa = mikeio.Dataset({"Foo": mikeio.DataArray([1, 2, 3])})
+    dsb = mikeio.Dataset({"Bar": mikeio.DataArray([4, 5, 6])})
+    with pytest.raises(ValueError):
```
jsmariegaard, line 1228: As comment above

mikeio/dataset/_dataset.py
ecomodeller: How to handle units in a pragmatic way is what I am struggling with, and I suppose one of the reasons that this is not already supported.

tests/test_dataset.py
ecomodeller: Let's try with minimal requirements. If two datasets have the same number of items, and the shapes of the data is identical, we can multiply them.

jsmariegaard: Additional thought: could this feature also be used for scaling several items in a file with the same spatially/temporally varying field? 

a) Should we allow that shapes conform but items in "other" is either the same as "this" or 1 (meaning same scaling applied to all variables).

b) Even further should we allow that as long as broadcasting rules allow multiplication then shape of "other" is okay? (so I could e.g. scale a dfsu with a dfs0)

ecomodeller: If we decide to go down this rabbit hole, I think we should consider restructuring MIKE IO to rely on xarray instead of numpy and to make use of the broadcasting in xarray. https://docs.xarray.dev/en/stable/user-guide/computation.html#broadcasting-by-dimension-name

jsmariegaard: Let's wait until someone has the need

PR #795, opened 2025-02-19
Add keyword arguments to interp_like instead of kwargs.

ecomodeller:


ecomodeller:
On second thought this PR might cause more problems, than it solves.

PR #804, opened 2025-03-18
Marimo notebooks

ecomodeller:


jsmariegaard:
Great to add these as a supplement but I think it is too early to get rid of the Jupyter equivalents. 

Question: what about the rendering online? I guess we would need to render them as part of building the documentation for people to see the outputs? Or can rely on an external service that could render on the fly?

ecomodeller:
> Great to add these as a supplement but I think it is too early to get rid of the Jupyter equivalents.
> 
> Question: what about the rendering online? I guess we would need to render them as part of building the documentation for people to see the outputs? Or can rely on an external service that could render on the fly?

If we embrace the interactive possibilities of marimo, exporting them to a static webpages is not ideal, but probably useful to get an impression if the content is relevant.🤔

jsmariegaard:
Could this approach by Vincent be adopted: https://youtu.be/W8nTSmZnQBs?si=lNGyj-T1X1QKK-S- ?

ecomodeller:
> Could this approach by Vincent be adopted: https://youtu.be/W8nTSmZnQBs?si=lNGyj-T1X1QKK-S- ?

Only if you can manage to convince someone to port `mikecore` to WASM.

Publishing a marimo notebook as a containerized client-server app with e.g. [Railway](https://railway.com/) is much easier, I made this example 6 months ago to test [Live app](https://marimo-test-production.up.railway.app/) [Source code](https://github.com/ecomodeller/marimo-test)

The downside is that it adds yet another service to pay for and to keep track of.

PR #808, opened 2025-03-25
Fix bug in Dfs0 read time slice

ecomodeller:
image


jsmariegaard (approved):
Thanks @ecomodeller! Looks great. 

Makes sense to deprecate old way of slicing!

PR #812, opened 2025-04-03
Access pfs with hierarchical path

ecomodeller:
Pfs files are structured in an hierarchical way, similar to file systems.

```python
# attributes
pfs.foo.bar.baz.simple
# like dictionary
pfs["foo"]["bar"]["baz"]["simple"]
```

This PR adds the possibility to use a nested path similar to a file system.

```python
path = "foo/bar/baz/simple"
pfs[path]
```

Similar to xml:
```python
import xml.etree.ElementTree as ET

xml_data = """<foo>
    <random>1</random>
    <bar>
        <speed>fast</speed>
        <baz>
            <simple>0</simple>
        </baz>
    </bar>
</foo>"""

root = ET.fromstring(xml_data)
root.find("bar/baz/simple").text
```


jsmariegaard (changes requested):
Functionality all looks good. Thanks

But shouldn't we let people know that this way of accessing nested keywords is possible- otherwise I don't think people are going to use it. Please add short description of this in the docs - maybe around here 

https://github.com/DHI/mikeio/blob/066393c283d2e96a89944c698d87684c434a83aa/docs/user-guide/pfs.qmd#L80

jsmariegaard (approved):
❤️ :shipit:

ecomodeller:
:shipit:

PR #816, opened 2025-04-04
Pfs set nested path

ecomodeller:
#812 added the ability to access parts of a PfsDocument using a nested path, e.g. 

```python
>>> pfs["foo/bar/baz"]
0
```

This PR extends the functionality to set values using a nested path.
```python
pfs["foo/bar/baz"] = 1
```


jpalm3r (commented):
Looks good. I just have two small suggestions to simplify the code.

PR #817, opened 2025-04-07
Remove support for list of numpy arrays from Dataset.__init__

ecomodeller:
Scheduled to be part of version 2.7.0


ecomodeller:
Part of #829 instead.

PR #830, opened 2025-05-23
Change dfs datatype

otzi5300:
Add generic function + test to change the datatype of a DFS:s file


mikeio/generic.py
```diff
@@ -959,3 +965,60 @@ def _get_repeated_items(
...
+    """
+    dfs_out = _clone(infilename, outfilename, datatype=datatype)
+    dfs_in = DfsFileFactory.DfsGenericOpen(infilename)
+
+    item_numbers = _valid_item_numbers(dfs_in.ItemInfo)
+    n_items = len(item_numbers)
+    n_time_steps = dfs_in.FileInfo.TimeAxis.NumberOfTimeSteps
+    deletevalue = dfs_in.FileInfo.DeleteValueFloat
+
+    # Write data to outputfile with new datatype
```
ecomodeller, line 1011: This can done in a simpler way, take a look at this example from mikecore: https://github.com/DHI/mikecore-python/blob/6ebbb1a1abebb88cd2060346611fa628eac57a82/tests/examples_misc.py#L59-L63

otzi5300: good, yes i had the feeling it could be simplified

tests/test_generic.py
```diff
@@ -638,3 +648,52 @@ def test_fill_corrupt_data(tmp_path: Path) -> None:
...
+
+    assert datatype_out == OUT_DATA_TYPE
+    assert n_timesteps_in == n_timesteps_out
+    # Also check that data is not modified
+    org = mikeio.read(infilename).to_numpy()
+    new = mikeio.read(outfilename).to_numpy()
+    assert np.allclose(org, new, rtol=1e-08, atol=1e-10, equal_nan=True)
+
+
+def test_change_datatype_dfsu(tmp_path: Path) -> None:
```
ecomodeller, line 677: I don't think it is a good idea to change the data type on dfsu, Mike Zero will refuse to open it. It is good enough with a test for dfs0.

ecomodeller (approved):

PR #833, opened 2025-05-28
Use TimeDeltaIndex for relative time

ecomodeller:
As an alternative to using a DateTimeIndex with epoch = 1970,1,1


ecomodeller:
Not relevant for most MIKE products, most engines use calendar axis.
image

PR #835, opened 2025-06-02
Interpolation ids and weights always go together

ecomodeller:
* New `Interpolant` class instead of tuple of ids and weights
* Avoid looping over timestep interpolation
* Move function interp2d to method onto the new Interpolant class

### DIFF
--- a/mikeio/dfsu/_dfsu.py
+++ b/mikeio/dfsu/_dfsu.py
@@ -214,8 +214,9 @@ class Dfsu2DH:
             )
         return str.join("\n", out)
 
+    # TODO change to GeometryFM2D
     @property
-    def geometry(self) -> GeometryFM2D:
+    def geometry(self) -> Any:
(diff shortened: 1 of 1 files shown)


jsmariegaard (approved):
I like the new class and the idea about keeping the weights and ids together. I did like that you could previously directly work on both ndarrays, datasets and dataarrays, but I guess you would like to avoid that from a typing perspective. 

When testing around I noticed some missing guardrails in the Dataset/DataArray.interp method: it has a n_nearest argument that only makes sense in specific cases. The method however silently accepts the argument even when it is not used or does not give meaning. That I do not like. I would prefer that it raises an error if n_nearest is supplied but not used. Maybe it should not be part of this PR, but it is pretty closely related...

It would be good to add more interpolation behaviour soon - e.g. from dfs2 also. Maybe next year.

PR #839, opened 2025-06-19
Generic derived variables

ecomodeller:
This PR adds the possibility to create derived variables without reading the entire file into memory.

The proposed `transform` function operates on data from a single timestep.

```python
import mikeio
from mikeio.generic import transform, DerivedItem
# calc current speed from u,v
derived = [
      DerivedItem(
          item=ItemInfo("Current Speed", mikeio.EUMType.Current_Speed),
          func=lambda x: np.sqrt(x["U velocity"] ** 2 + x["V velocity"] ** 2),
      ),
 # add more items, e.g. direction
]
# include existing items except Total Water Depth
infilename = "HD.dfsu"
dfs1 = mikeio.Dfsu2DH(infilename)
sel_items = [
      DerivedItem(item=item)
      for item in dfs1.items
      if item.name != "Total Water Depth"
]
sel_items.extend(derived)

# Transform
transform(infilename, "transformed.dfsu", vars=sel_items)
```

The default is to append derived items to the list of existing items, controlled by the `keep_existing_items` flag.

### DIFF
--- a/mikeio/generic.py
+++ b/mikeio/generic.py
@@ -1043,7 +1043,7 @@ class DerivedItem:
     def __init__(
         self,
         name: str,
-        type: EUMType,
+        type: EUMType | None = None,
         unit: EUMUnit | None = None,
         func: Callable[[Mapping[str, np.ndarray]], np.ndarray] | None = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


mikeio/generic.py
```diff
@@ -1011,3 +1012,127 @@ def change_datatype(
...
+
+    dfs = _clone(
+        str(infilename),
+        str(outfilename),
+        items=items,
+    )
+
+    n_time_steps = dfs_i.FileInfo.TimeAxis.NumberOfTimeSteps
+
+    for timestep in range(n_time_steps):
```
ryan-kipawa, line 1116: Have you checked performance on files with many time steps?

ryan-kipawa (approved):
Looks like an easy to use syntax, nice documentation, and great typing. My only comment is in regards to whether files with many timesteps are performant. That could also be an improvement later on (if it's an issue)...

mikeio/generic.py
ecomodeller: Nope. But this is the same type of loop we have for dfs files except dfs0. I.e. I would not use use this function to transform dfs0 files, which generally fits in memory.

PR #847, opened 2025-10-08
Concise element coordinate calculation

ecomodeller:


ecomodeller:
@ryan-kipawa Could you do an independent review of this PR? Considering all options of 2d, 3d, 2d vertical, triangles and quads.

Issue #68, opened 2020-08-26
Select/filter Dataset on time and/or items


jsmariegaard:
If I have a Dataset from a dfsu file, I can very conveniently filter on element_ids and save to new file. That's great!

    selds = ds.isel(idx=elem_ids)
    dfs.write(newfile, selds, element_ids=elem_ids)

I would like to similarly filter on timesteps and/or items as a method on the Dataset class returning a new Dataset. Something like this:

     selds = ds.select_steps(range(4,19)) 

or 

     selds = ds.select_steps('2018-1-1','2018-2-1')

or similar with a daterange or something. And likewise on items either by name or number. These selections can then be chained: 

    selds = ds.select_items('water level').select_steps('2018-1-1','2018-2-1').isel(element_ids)

Or in a single method like this:

    selds = ds.select(items=['water level'], time_steps=['2018-1-1','2018-2-1'], element_ids=element_ids)

What do you think?

ecomodeller:
I think we should try to mimic the functionality of xarray http://xarray.pydata.org/en/stable/generated/xarray.Dataset.sel.html

Since we are not loading the data lazily (yet, at least) the filtering is most efficient to do while reading the data and not on the Dataset, but this should be possible, for convenience.

ecomodeller:
```
Python 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32

>>> import mikeio
>>> ds = mikeio.read("tests/testdata/HD2D.dfsu")
>>> ds
DataSet(data, time, items)
Number of items: 4
Shape: (9, 884)
1985-08-06 07:00:00 - 1985-08-07 03:00:00
>>> ds.items
[Surface elevation <Surface Elevation> (meter), U velocity <u velocity component> (meter per sec), V velocity <v velocity component> (meter per sec), Current speed <Current Speed> (meter per sec)]
>>> ds2 = ds[['Surface elevation','Current speed']] # item selection
>>> ds2.items
[Surface elevation <Surface Elevation> (meter), Current speed <Current Speed> (meter per sec)]
>>> len(ds2)
2
>>> ds3 = ds2.isel([0,1,2], axis=0) # temporal selection
>>> ds3
DataSet(data, time, items)
Number of items: 2
Shape: (3, 884)
1985-08-06 07:00:00 - 1985-08-06 12:00:00
>>> ds4 = ds3.isel([100,200], axis=1) # element selection
[...]

PR #874, opened 2025-12-09
Add lazy evaluation API for memory-efficient DFS file operations

ecomodeller:
## Summary

Implements a Polars-inspired lazy evaluation API for composable, memory-efficient operations on large DFS files. 

**New module:** `mikeio.lazy`

**API:**
```python
from mikeio.lazy import scan_dfs

(scan_dfs("large_file.dfsu")
    .select(["Temperature", "Salinity"])
    .filter(time=slice("2020-01-01", "2020-12-31"))
    .rolling(window=24, stat="mean")
    .with_items(Temperature=lambda x: x * 1.8 + 32)
    .to_dfs("output.dfsu")
)
```

**Operations:**
- `.select()` - Filter items by name or index
- `.filter()` - Filter timesteps with slice notation
- `.aggregate()` - Temporal aggregation with multi-stat support and customizable naming
  - Single pass multi-aggregation: `.aggregate(["min", "max", "mean", "std"])`
  - Custom labels: `.aggregate(["mean"], labels={"mean": "Avg"})`
  - Custom format: `.aggregate(["max"], format="{item} {stat}")`
- `.rolling()` - Rolling statistics (mean, min, max, median, sum, std) + custom functions
- `.with_items()` - Transform items with custom functions
- `.scale()` - Scale all items with factor and offset
- `.diff()` - Compute difference with another file
- `.to_dfs()` - Execute pipeline and write output

**Key features:**
- Single-pass execution - avoids intermediate files
- Multi-aggregation - compute multiple statistics efficiently in one pass (like DfsuStatistics.exe)
- Customizable naming - control output item names with labels and format strings
- Memory-efficient - processes data in chunks/windows with configurable buffer sizes
- Composable - chain operations in any order
- Lazy evaluation - operations queued until `.to_dfs()` is called
- Validation - operations validated against file metadata before execution

**Design principles:**
- Internal consistency: Uses `.with_items()` and `.to_dfs()` to match MIKE IO conventions
- External inspiration: Draws from Polars for composability and lazy evaluation patterns
- Reuses utilities from `generic.py` for robustness

## Relation to #68

Issue #68 requested chainable selection methods and noted that lazy filtering while reading would be more efficient. This PR provides that capability for files too large to load into memory, complementing the existing in-memory Dataset operations.

### DIFF
--- a/src/mikeio/lazy.py
+++ b/src/mikeio/lazy.py
@@ -42,7 +42,7 @@ from mikecore.DfsFileFactory import DfsFileFactory
 
 # Import utilities from generic module
 from .eum import ItemInfo
-from .generic import _TimeInfo, _clone, _valid_item_numbers
+from .generic import DerivedItem, _TimeInfo, _clone, _valid_item_numbers
 
 __all__ = ["scan_dfs", "LazyDfs"]
 
@@ -97,6 +97,7 @@ class ExecutionContext:
     rolling: "RollingOp | None" = None
     aggregate: "AggregateOp | None" = None
     diff_file: Path | None = None
+    derived_items: list[DerivedItem] | None = None
 
 
 # ============================================================================
@@ -433,6 +434,28 @@ class DiffOp(Operation):
     def apply(self, context: ExecutionContext) -> None:
         """Store diff config for later application."""
         context.diff_file = self.other_file
+
+
+class DeriveOp(Operation):
+    """Operation to derive new items from existing items."""
+
+    def __init__(self, derived_items: list[DerivedItem]):
+        self.derived_items = derived_items
+
+    @property
+    def affects_metadata(self) -> bool:
+        return True  # Adds new items to the output
+
+    def name(self) -> str:
+        return "derive"
+
+    def explain(self) -> str:
+        item_names = ", ".join(item.item.name for item in self.derived_items)
+        return f"DERIVE new items: [{item_names}]"
+
+    def apply(self, context: ExecutionContext) -> None:
+        """Store derived items for later application."""
+        context.derived_items = self.derived_items
 
 
 # ============================================================================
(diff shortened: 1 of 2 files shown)


src/mikeio/lazy.py
```diff
@@ -0,0 +1,1168 @@
...
+
+        Examples
+        --------
+        >>> # Compare model run with observations
+        >>> (scan_dfs("model_results.dfsu")
+        ...     .diff("observations.dfsu")
+        ...     .to_dfs("model_error.dfsu")
+        ... )
+
+        >>> # Compare two model scenarios
```
ecomodeller: These examples are not executed. E.g. using ```{python. How do ensure the docstring remain correct?

ecomodeller:
**Random ideas:**
* Progress bar
* Other nice to have functionalities available in `Dataset`
  - `.rename()`
* Support multiple aggregations stored in a single output file
  - Suffix variable to indicate aggregation

src/mikeio/lazy.py
```diff
@@ -0,0 +1,1168 @@
...
+            return
+
+        # Get data shape from first item to aggregate
+        first_itemdata = context.dfs_in.ReadItemTimeStep(
+            item_numbers_to_aggregate[0] + 1, start_step
+        )
+        n_elements = len(first_itemdata.Data)
+        n_timesteps = end_step - start_step
+
+        # Create buffers: [n_items][n_timesteps, n_elements]
```
ecomodeller: Allocating these arrays defies the entire purpose of the `lazy` module (files larger than memory). This has to be done in a different way. 🤬

ryan-kipawa (commented):
For custom labels and formatting: it would be helpful to see an example of its output in the docs

src/mikeio/lazy.py
```diff
@@ -0,0 +1,1740 @@
...
+        This method triggers the actual execution of all queued operations.
+        The file is processed efficiently in a single pass.
+
+        Parameters
+        ----------
+        path : str | Path
+            Output file path
+        buffer_size : float, optional
+            Maximum memory buffer size in bytes for processing large files,
+            default 1e9 (1 GB)
```
ryan-kipawa: How does a user know what to set this to? I imagine most will use the default, but you need to have some special expertise to set it properly. For example, if I have 32GB of memory available, can I get more performance by putting this as 32 GB? If we know such rules exist, it could be an option to dynamically select one we think is most performant as well.

src/mikeio/lazy.py
```diff
@@ -0,0 +1,1740 @@
...
+    def to_dfs(self, path: str | Path, buffer_size: float = 1e9) -> None:
+        """Execute pipeline and write to DFS file.
+
+        This method triggers the actual execution of all queued operations.
+        The file is processed efficiently in a single pass.
+
+        Parameters
+        ----------
+        path : str | Path
+            Output file path
```
ryan-kipawa: Will it overwrite by default? Will it create non-existing parent directories?

src/mikeio/lazy.py
```diff
@@ -0,0 +1,1740 @@
...
+            stat_label = context.aggregate.get_stat_labels()[0]
+            items_list = []
+
+            # Get original item info for selected items
+            # Skip Z coordinate for layered dfsu
+            is_layered = (
+                len(context.item_numbers) > 0
+                and context.dfs_in.ItemInfo[context.item_numbers[0]].Name
+                == "Z coordinate"
+            )
```
ryan-kipawa: Now I am convinced this belongs in a dfs module somewhere 😛

ryan-kipawa (commented):
General dicsussion point:

The operation classes feel a bit superficial. Would it make sense for LazyDfs to hold the execution context, and put the logic of those execution classes directly in the methods on LazyDfs?

src/mikeio/lazy.py
```diff
@@ -0,0 +1,1740 @@
...
+        # Check for aggregate operation
+        if context.aggregate is not None:
+            self._process_with_aggregate(context, context.aggregate)
+            return
+
+        # Check for rolling operation
+        if context.rolling is not None:
+            self._process_with_rolling(context, context.rolling)
+        else:
+            self._process_simple(context)
```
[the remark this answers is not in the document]
ecomodeller: Agree, it would be really great if we could manage to create a complete lazy api.

PR #877, opened 2025-12-10
Refactor IDW interpolation: simplify logic and clarify API

ecomodeller:
## Summary

- Simplified IDW interpolation function by removing unnecessary conditional logic
- Clarified public vs private API boundaries
- Updated documentation and notebooks to use self-contained examples instead of private functions

## Why this matters

Documentation that imports private functions teaches users to depend on implementation details. When users copy examples from docs, those imports become de facto public API—making internal refactoring a breaking change. This undermines the purpose of private/public separation and makes version upgrades break user code.

### Code simplification
- Renamed `get_idw_interpolant` → `_get_idw_interpolant` (make private)
- Removed 1D/2D conversion logic (function now requires 2D input)
- Cleaner normalization using `keepdims=True`
- Added documentation explaining assumptions (distances must be sorted)

### Testing
- Removed direct unit test of private function
- Behavior still validated through integration tests:
  - `test_interp2d_same_points` verifies exact match handling
  - Other tests verify full interpolation pipeline

### Documentation
- Updated scatter interpolation example with copy-pasteable `simple_idw()` function
- Removed this section from Jupyter notebook with same approach
- Removed reliance on private mikeio functions
- Makes it clear users should implement their own solution for this use case


ryan-kipawa (changes requested):
I agree with the overall principles of this PR. However, I suspect this is very common functionality, and since it's been in the documentation for so long, it could be argued it's already part of the public API despite having a leading underscore. How do you see that?

We could consider

1. Support such functionality in the public API
2. If we remove it, keep _get_idw_interpolate with a deprecation warning that it will be removed and is not part of the public API

PR #888, opened 2026-01-08
Add title attribute to Dataset and DFS file readers

### LINKED ISSUE #883: Ellaborate kwargs for method .to_dfs() on Dataset and DataArray classes
Hi everyone

First of all, I figured out my problem. And I would like to share this with you all, and hopefully it will be a little easier to do in the future ;)

Anyways, my (small) problem was being able to set a title for dfs-files (in this case `.dfs0`). In the manual way, you would open a `dfs`-file in MIKE ZERO and add a title in the "File properties" window. But I figured it would be nicer and more quick to do using `mikeio`.

So, first of all, I thougt that I just have to add a `title` attribute for my `DataArray` instance. However, the procedure was to add the `DataArray` instance to a `Dataset` instance and then use the method `.to_dfs(filename, **kwargs)`. Actually, the same method is available for the DataArray class. But the two methods differ a little when it comes to the keyword arguments (`**kwargs`). Because for the `Dataset` class, a keyword argument can be `title`, which is not the case for the `DataArray` class.
Thus, when I call the method `.to_dfs(filename='test.dfs0',title="Test")` for my `Dataset` instance, I actually get a title "Test" in the output. Happy days!
Similar method and input arguments for a `DataArray` instance, a title is not added in the output, even though this makes sense, since it is the overall dataset I want a title for.

It is not to criticize this module or its contents, because I actually love it! I only suggest that you ellaborate the keyword arguments for both classes, `Dataset` and `DataArray`. I don't want you to drop `**kwargs` argument, just describe what arguments that are covered.

Thanks in advance!

otzi5300:
DFS files have a title metadata field displayed in [MIKE Zero](https://www.dhigroup.com/technologies/mikepoweredbydhi/mike-zero). Previously this was ignored on read and lost on write. Resolves user request for title preservation through processing pipelines.

- `Dataset.title` attribute, preserved through all operations (sel, isel, aggregation, arithmetic, concat, merge, etc.)
- `title` property on all file readers (Dfs0–Dfs3, Dfsu2DH, DfsuLayered, DfsuSpectral)
- `to_dfs()` writes `Dataset.title` to the file
- `to_xarray()` propagates title as an `attrs["title"]`

```python
ds = mikeio.read("file.dfs0")
ds.title  # read from file
ds.title = "My Title"
ds.to_dfs("output.dfs0")  # written to file
```

### DIFF
--- a/docs/user-guide/dfs0.qmd
+++ b/docs/user-guide/dfs0.qmd
@@ -92,6 +92,17 @@ ds_typed
 ```
 
 
+## File title
+
+DFS files have a title field shown e.g. in MIKE Zero. The title is read from the file and preserved through processing. You can set it before writing:
+
+```{python}
+ds = mikeio.read("mauna_loa_co2.dfs0")
+ds.title = "Mauna Loa CO2"
+ds.to_dfs("mauna_loa_co2.dfs0")
+```
+
+
 ## Accumulated datavalue type
 
 Some dfs0 items use an accumulated data value type rather than the default instantaneous type. This is common for precipitation data and can be specified using the `data_value_type` argument of [](`mikeio.ItemInfo`). Refer to the [MIKE documentation](https://doc.mikepoweredbydhi.help/webhelp/2026/timeserieseditor/TSEdit/TSEdit/TS_Types_graphical_representation.htm) for guidance on which type to use for your setup.
--- a/src/mikeio/dataset/_dataarray.py
+++ b/src/mikeio/dataset/_dataarray.py
@@ -1837,20 +1837,18 @@ class DataArray:
             {self.name: self}
         )  # Single-item Dataset (All info is contained in the DataArray, no need for additional info)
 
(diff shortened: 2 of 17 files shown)


ecomodeller (changes requested):
The implementation works but has two API design issues:

**1. `title` parameter not discoverable in `to_dfs()`**

The `title` kwarg is hidden behind `**kwargs`. Users won't find it via autocomplete. Make it explicit:

```python
def to_dfs(self, filename: str | Path, title: str = "", **kwargs: Any) -> None:
```

Applies to both `Dataset.to_dfs()` and `DataArray.to_dfs()`.

**2. Title lost in read/write round-trips**

Dataset has no `title` attribute, so this loses the title:

```python
ds = mikeio.read("file.dfs0")
ds.to_dfs("output.dfs0")  # title becomes ""
```

Consider adding `title` to Dataset so it's preserved automatically.

ecomodeller (approved):

PR #922, opened 2026-02-21
Enable ruff EM rules for cleaner tracebacks

ecomodeller:
## Summary

Enables the `EM` (error message) ruff rules and fixes all 187 violations across the codebase.

### Why?

When a string literal is passed directly to an exception, Python's traceback shows the message twice — once in the source context and once as the exception value:

```
Traceback (most recent call last):
  File "example.py", line 3, in <func>
    raise ValueError(f"Item not found: {name}")
ValueError: Item not found: foo
```

By assigning the message to a variable first, the traceback becomes cleaner:

```
Traceback (most recent call last):
  File "example.py", line 4, in <func>
    raise ValueError(msg)
ValueError: Item not found: foo
```

The error message appears exactly once, making tracebacks easier to scan — especially for long messages or deeply nested stacks. This is enforced going forward by the `EM101` (string literals) and `EM102` (f-string literals) lint rules.


ecomodeller:
Not useful enough.

PR #936, opened 2026-03-15
Rename axis to dim on aggregation methods

ecomodeller:
## Summary
- Rename `axis` parameter to `dim` on all DataArray/Dataset aggregation methods (mean, max, min, std, quantile, etc.) to align with xarray's API
- Old `axis=` keyword preserved as deprecated keyword-only arg with `FutureWarning`
- All tests migrated to use `dim=`


ecomodeller:
Not right now

PR #944, opened 2026-03-21
Document unweighted spatial mean, add mean vs average tests

ecomodeller:
## Why

mean(axis="space") on unstructured meshes is unweighted — all elements contribute equally regardless of area. On HD2D.dfsu, where the largest element is 15x the smallest, this gives biased results. Users must use average(weights=area) for correct area-weighted means, but nothing in the API or docstrings warns about this.

## What this adds

- 4 tests demonstrating the difference on variable meshes, agreement on uniform grids, and a synthetic known-weights case
- Updates mean() and nanmean() docstrings to note they are unweighted and reference average()

### DIFF
--- a/src/mikeio/dataset/_dataarray.py
+++ b/src/mikeio/dataset/_dataarray.py
@@ -1331,8 +1331,8 @@ class DataArray:
 
         Note: on unstructured meshes with variable element sizes,
         ``mean(axis="space")`` treats all elements equally regardless
-        of area. For an area-weighted spatial mean, use
-        :meth:`average` with ``weights=geometry.get_element_area()``.
+        of area. For an area-weighted spatial mean, use ``average()``
+        with ``weights=geometry.get_element_area()``.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


src/mikeio/dataset/_dataarray.py
```diff
@@ -1327,7 +1327,12 @@ def min(self, axis: int | str = 0, **kwargs: Any) -> DataArray:
         return self.aggregate(axis=axis, func=np.min, **kwargs)
 
     def mean(self, axis: int | str | None = 0, **kwargs: Any) -> DataArray:
-        """Mean value along an axis.
+        """Unweighted mean value along an axis.
```
ecomodeller, line 1330: @otzi5300 I was about to add this clarification. Based on the discussion related to vertical mean, maybe the semantics of `mean` should change and take area/volume into account, instead of doing the naive mean?

But thinking about this for 10 sec, it seems safer to add a new method or argument to do weighted aggregations. #979

PR #950, opened 2026-03-27
ADR-010: No generic DataArray[T]/Dataset[T]

ecomodeller:
## Summary

Documents the decision not to make `DataArray` and `Dataset` generic over geometry type `T`.

The generic type parameter would only be useful in the narrow window between reading a file and the first spatial operation (`isel`/`sel`/aggregation), since those operations change the geometry type at runtime. Making the type parameter useful beyond that would require either complex overload chains or changing `isel`/`sel` semantics to preserve geometry type — both of which add complexity that serves the type system rather than the user.


ecomodeller:
@copilot fix numbering of the ADR

PR #977, opened 2026-05-19
Add public z-coordinate accessor for layered dfsu DataArrays

### LINKED ISSUE #976: Expose time-varying node z-coordinates as a public API on 3D dfsu DataArrays
For 3D layered dfsu files, node z-coordinates move at every timestep (free surface, sigma stretching). MIKE IO stores this as a `(n_time, n_nodes)` array but exposes it only as the private `DataArray._zn` / `Dataset._zn`.

## Why it matters

Time-varying z's are a first-class physical quantity for 3D work:

- **Oceanography:** pycnocline / mixed-layer depth, buoyancy frequency — need `z(t)` alongside `T(t), S(t)`.
- **Sediment / spill:** vertical flux through a depth needs the actual layer-interface elevation per timestep, not the static mesh z.
- **Ecosystem / water quality:** light attenuation and depth-integrated production change with the tide.
- **Layer-weighted stats:** depth-averaged velocity and volume-weighted means need `dz(t) = diff(zn(t))` along a column.
- **Re-gridding to fixed z-levels** for comparison with CTD casts, moorings, or [CF-compliant](https://cf-convention.github.io/) NetCDF.

The project already relies on `_zn` in places users can see:

- The [Dfsu → netCDF export notebook](https://github.com/DHI/mikeio/blob/main/notebooks/Dfsu%20-%20Export%20to%20netcdf.ipynb) uses `ds._zn` to build the dynamic z-coordinate.
- Vertical-profile plotting (`src/mikeio/dataset/_data_plot.py:638`) and the 3D writer (`src/mikeio/dfsu/_dfsu.py:125`) read `_zn` directly.

So the data is loaded and routinely needed; it's just hidden behind a private name.

## Proposal

Add a public accessor. Open questions:

1. Raw ndarray, or wrapped with time/geometry?
2. Node-based, element-center, or both?
3. Name — `zn`, `node_z`, `z_coordinates`?

ecomodeller:
Closes [#976](https://github.com/DHI/mikeio/issues/976).

## Why

Users of layered 3D dfsu files need the time-varying node z-coordinates to do anything depth-aware — vertical interpolation, dz-weighted column means, CF-conformant NetCDF export — but today the only way to get them is `da._zn`, a private attribute. The issue ([#976](https://github.com/DHI/mikeio/issues/976)) tracks the request to make this a proper part of the public API. The private-attribute path is already in use in third-party tooling and in our own example notebooks, which is the strongest signal that the surface needs to be public.

## Before / After

```python
import mikeio
ds = mikeio.read("oresund_sigma_z.dfsu")
da = ds["Temperature"]

# Before — reach into a private attribute
zn = da._zn                                    # (n_time, n_nodes)
ze = np.array([                                # element centers had to be derived by hand
    da._zn[:, np.asarray(nodes)].mean(axis=1)
    for nodes in da.geometry.element_table
]).T

# After — public namespace, dispatched on geometry
zn = da.z.nodes                                # (n_time, n_nodes)
ze = da.z.elements                             # (n_time, n_elements), lazily computed and cached

# Non-layered DataArrays fail discoverably instead of returning silent None
mikeio.read("random.dfs2")[0].z.nodes
# AttributeError: DataArray with geometry 'Grid2D' has no z-coordinates;
#                 only layered 3D dfsu DataArrays expose .z
```

## What

A new namespace accessor `da.z` on `DataArray` (and a delegating `ds.z` on `Dataset`), dispatched on `type(self.geometry)` exactly like the existing `da.plot` namespace. For layered geometries `da.z.nodes` returns the per-timestep node z-coordinates (same array as `_zn`) and `da.z.elements` returns the per-element mean of column-node z's, computed lazily and cached.

In-tree consumers migrated to the public surface:

- Vertical-profile plotter (`_data_plot.py`)
- 3D dfsu writer (`_dfsu.py`)
- Layered surface-elevation extractor (`_layered.py`)
- [Dfsu → netCDF export notebook](https://github.com/DHI/mikeio/blob/main/notebooks/Dfsu%20-%20Export%20to%20netcdf.ipynb)
- Existing `_zn` test assertions in `test_dfsu3d.py` and `test_dataset.py`

A user-guide section in `docs/user-guide/dfsu.qmd` documents `nodes` / `elements` with a worked dz-weighted column-mean example.

## Design

The choice between accessor-namespace vs. subclass vs. free function vs. geometry method is laid out in [`specs/issue-976-z-coordinates-public-api/tradeoff.md`](https://github.com/DHI/mikeio/blob/feat/issue-976-z-accessor/specs/issue-976-z-coordinates-public-api/tradeoff.md). The accessor pattern was chosen because it reuses an established in-tree convention (`da.plot`), keeps `DataArray` a single class ([ADR-003](https://github.com/DHI/mikeio/blob/main/adr/003-dataset-dataarray-pattern.md)), keeps geometry static ([ADR-004](https://github.com/DHI/mikeio/blob/main/adr/004-geometry-abstraction.md)), and avoids the six-site `return DataArray(...)` refactor a subclass would force.

## Scope

`_zn` is kept as a private alias for one release cycle — this PR does not remove it. A setter / write path, a time-varying geometry abstraction, and CF-conformant NetCDF helpers are all explicitly out of scope here.

## A note on `specs/`

This PR also adds `specs/issue-976-z-coordinates-public-api/` — output from the AndThen workflow used to design the feature. The FIS and the requirements-clarification doc are one-shot workflow artefacts and will go stale (they reference specific line numbers), so the plan is to delete them after merge. The [`tradeoff.md`](https://github.com/DHI/mikeio/blob/feat/issue-976-z-accessor/specs/issue-976-z-coordinates-public-api/tradeoff.md) is genuinely useful for future readers asking "why an accessor and not a subclass?" — I'd suggest promoting it into a new `adr/005-z-coordinate-accessor.md` alongside ADR-003 and ADR-004 it cites, and dropping the rest of `specs/` in a follow-up commit before merge. Happy to do that on this branch if reviewers agree.


src/mikeio/dataset/_z_accessor.py
```diff
@@ -0,0 +1,122 @@
...
+    def _compute_elements(self) -> np.ndarray:
+        zn = self.nodes
+        element_table = self._da.geometry.element_table
+        n_elements = len(element_table)
+
+        if zn.ndim == 2:
+            n_time = zn.shape[0]
+            ze = np.empty((n_time, n_elements), dtype=zn.dtype)
+            for j, nodes in enumerate(element_table):
+                ze[:, j] = zn[:, np.asarray(nodes, dtype=int)].mean(axis=1)
```
otzi5300: I think we can use nodes straight away
ze[:, j] = zn[:, nodes].mean(axis=1)

but maybe better to be defensive?

otzi5300 (approved):

src/mikeio/dataset/_z_accessor.py
ecomodeller: Simpler is better!

PR #982, opened 2026-06-01
Fix crash reading dfs0 with an item named like a reserved attribute

ecomodeller:
`mikeio.read()` raises `AttributeError: property 'z' of 'Dataset' object has no setter` on any dfs0 whose vertical-coordinate item is literally named `z`.

### Why

[#977](https://github.com/DHI/mikeio/pull/977) added a read-only `z` property to `Dataset`. But `Dataset` also exposes every item as a dynamic instance attribute (`ds.Surface_Elevation`) by doing `setattr(self, safe_name, da)` during construction. An item named `z` therefore tries to assign to the setter-less `z` property and crashes — before any data can be returned.

This was caught by modelskill's ["test against mikeio main" canary](https://github.com/DHI/modelskill/actions/runs/26739203873/job/78798923746) (all failures are in its *vertical* tests, which read dfs0 files carrying a `z` item). It is entirely within mikeio — released mikeio is unaffected.

`z` is just the first one hit: the same `setattr` collision crashes for any item named after a setter-less property (`geometry`, `shape`, `values`, …) and *silently shadows* methods like `mean`/`max`, so `ds.mean()` would break for a file with a `mean` item.

### Fix

Skip the convenience attribute whenever an item's safe-name matches a class member (property or method). The item stays reachable via `ds["name"]`; `ds.<name>` keeps meaning the reserved member — the same `[]`-vs-`.` rule that already applies to `time` and `geometry`. This resolves the whole family of collisions in one place rather than patching `z` alone.

Behavior change worth flagging: for an item whose name matches a reserved member, `ds.<name>` now returns the method/property instead of the item. The item is always available via `ds["<name>"]`.

---

Also adds [ADR-009](adr/009-item-attribute-access.md), recording why item attribute access exists (pandas/xarray lineage), the reserved-name collision it causes, and the decision behind this fix (collisions made index-only and silent). It frames attribute access as **frozen legacy** — `ds["name"]` is canonical — with deprecation toward removal under evaluation for v4 (a separate follow-up; no doc or API changes here).

### DIFF
--- a/adr/009-item-attribute-access.md
+++ b/adr/009-item-attribute-access.md
@@ -1,0 +1,92 @@
+# ADR-009: Item Attribute Access on Dataset
+
+**Status:** Accepted — attribute access frozen; `ds["name"]` is canonical; full deprecation under evaluation for v4
+**Date:** 2026-06
+
+## Context
+
+A `Dataset` is a name-keyed collection of `DataArray` items (see
+[ADR-003](003-dataset-dataarray-pattern.md)). The canonical way to select an item
+is by key: `ds["Surface elevation"]`.
+
+Following the [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html)
+/ [xarray](https://docs.xarray.dev/en/stable/user-guide/data-structures.html#dictionary-like-methods)
+convention that MIKE IO's data structures were modelled on, items are *also*
+exposed as instance attributes for ergonomics and tab-completion:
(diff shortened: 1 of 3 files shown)


ryan-kipawa (commented):
Looks okay 👌 

Only one comment worth mentioning:


- The ADR rationale seems a bit logically inconsistent, and perhaps too narrow. My assumption is the rationale is actually: we prefer explicitly typed APIs, like Polars, even if it means deviating from Pandas / Xarray conventions.

PR #983, opened 2026-06-03
Feat/issue 976 z accessor

otzi5300:
Added vertical mean, max, min

user can now do ds.max(axis="z) or da.max(axis="z) 

or if they want to subset first

da.sel(layers=[0,1].max(axis="z) gives maximum of the two deepest layers

**test**
Two test files added:
- On simple with 3 sigma and 2 z. No sigma stretching
- On more complex with 3 sigma and 2 z. On second timesteps, surface level is increase and sigma streched. Bottom layer in element 2 is nan


otzi5300:
Commit to wrong branch and open PR. Closing PR

PR #1007, opened 2026-08-17
Expose dfs2 custom blocks as Dataset.custom_blocks

### LINKED ISSUE #283: Defining landvalue when writing dfs2-file
There does not seem to be an option to define a landvalue when wrtiting a dfs2-file from python with Dfs2().write. The landvalue is greyed out when opening the file in MIKE Zero. 
We neeed the functionallity when importing a bathymetry into a simulation-setup such as .m21fm. When the dfs2 is used as a domain/mesh we are prompted with the error: "no landvalue set". How can we specify a custom landvalue?

ryan-kipawa:
Closes #283. Supersedes #675.

Custom blocks are named arrays of metadata in the dfs header. MIKE 21 keeps the **land value** in `M21_Misc[3]`; without it MIKE Zero greys the land value out and an `.m21fm` setup rejects the bathymetry. MIKE IO could not read them at all, and `mikeio.read(f).to_dfs(g)` destroyed what `mikeio.generic.scale(f, g)` has always preserved.

## `Dataset.custom_blocks`

```python
ds = mikeio.read("waves.dfs2")
ds.custom_blocks
# {'M21_Misc': array([-1.77e-05, 0., 4.51, 10., 0., 0., 0.], dtype=float32)}

ds.custom_blocks["M21_Misc"][3] = -10.0        # MIKE 21 land value
ds.to_dfs("waves_with_land_value.dfs2")
```

A plain `dict[str, np.ndarray]`, editable in place or replaced wholesale. A list is read as `float32`; pass an array for any other dtype:

```python
ds.custom_blocks = {"M21_Misc": [0, 0, -900, -10, 0, 0, 0]}
ds.custom_blocks["Counts"] = np.array([1, 2, 3], dtype=np.int32)
```

Blocks are copied on assignment, carried through Dataset operations (`isel`, `sel`, `interp`, `interp_like`, aggregations, arithmetic, item selection), and written for dfs0, dfs1, dfs2 and dfs3.

## `Dfs0.custom_blocks`, `Dfs1`, `Dfs2`, `Dfs3`

Read-only, and read from the header alone — so the land value of a large bathymetry can be inspected without reading its data:

```python
mikeio.Dfs2("bathy.dfs2").custom_blocks["M21_Misc"][3]
```

## Behaviour worth knowing

- A Dataset stores whatever is assigned; the dfs rules (1-D, C-contiguous, one of the seven dfs dtypes, non-empty ASCII name) are applied on write, before the file is created.
- Propagation is blind: `isel(y=0)` writes a dfs2 `M21_Misc` verbatim into the resulting dfs1. Whether a block still means anything after the geometry changes is the caller's call.
- dfsu ignores them — its sole block, `MIKE_FM`, is regenerated from the geometry on every write. `DataArray.to_dfs` writes none either, having no file-level metadata.
- Reading is lenient: a block MIKE IO would refuse to write is skipped with a warning, so no file becomes unreadable.
- An `int16` block reads back as `uint16` (mikecore reads `Short` as `c_uint16`); prefer `int32`. Writing is correct.

## Docs

`user-guide/dfs2.qmd` gains a **Custom blocks** section with the `M21_Misc` layout; `examples/dfs2/bathy.qmd` gains a land-value step on the GEBCO→dfs2 walkthrough.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ryan-kipawa:
Pushed 8a92cfd9, addressing a manual review and verification pass against the MIKE SDK behaviour.

**Preservation stays dumb, on purpose.** Blocks are now carried through every operation that returns a `Dataset`, unconditionally — mirroring `generic._clone`, which copies a dfs header's custom blocks verbatim regardless of what else changes. The previous "drop when the geometry type changes" rule was MIKE IO inventing mutation semantics it has no business owning. A block can now end up describing a geometry it no longer matches; the user modifies or clears it, and the docs say so.

**Scope widened to dfs0/dfs1/dfs2/dfs3.** All four read and write custom blocks, each file object exposing a `custom_blocks` property. dfsu remains the exception: `MIKE_FM` is generated from the geometry by `mikecore`'s builder, which offers no way to add others — confirmed it is MIKE's to write and not the user's to modify — so a dfsu `Dataset` reports no blocks and writing one that carries them drops them with a warning. (PR title is now narrower than the feature; happy to rename.)

**`M21_Misc` is float32, not float64.** A plain sequence carries no dtype and numpy's default was wrong for the one block most users will ever write, so sequences now coerce to `float32` and only an explicit ndarray picks another dtype. Semantics confirmed as seven values: `{orientation (matching the projection), drying depth, -900 = has projection, land value, 0, 0, 0}`, per `DfsCustomBlock<float>.Create("M21_Misc", ...)`.

**No shared memory with the DFS API, and no way to fake it.** The out-of-sync scenario is real in principle but unreachable here: MIKE IO reads into memory and closes the file handle immediately, so there is nothing left to stay in sync with — numpy views over dfs-owned memory would dangle. `Dfs0/1/2/3.custom_blocks` therefore return a fresh copy on every access, so mutating a file object's blocks cannot look like it wrote something. The `Dataset` dict is the only writable path, and the round-trip tests read back from the written file rather than trusting the in-memory dict.

**Docs and tests trimmed.** The `custom_blocks` docstring is back to house length — what the blocks are is MIKE's documentation, not ours — with the table and caveats living in the dfs2 user guide, plus examples for editing the `M21_Misc` land value in place and adding a new block. `_warn_custom_blocks_not_written` was too shallow for one call site and is inlined into `write_dfsu`. The 15 per-operation preservation tests collapse into one test with a few representative asserts: the tests should state the intent of the API, not exhaustively prove it. The "not in repr" test is gone.

ryan-kipawa:
This did not turn out to be a one-shotted feature, at least not in a clean way. Anyways, here it is 😄

PR #1038, opened 2026-09-01
Encapsulate attributes that silently corrupted derived state

ecomodeller:
Several public attributes could be assigned or edited in place with no error, while values derived from them were cached or mirrored elsewhere. The result was a wrong number rather than an exception — the worst outcome for the many MIKE IO users who are engineers first and Python programmers second.

What could go wrong before this change:

- `da.time = pd.date_range(...)` with the wrong number of timestamps was accepted (the constructor checks it, assignment did not) and `to_dfs` then wrote a silently truncated file.
- `da.geometry = ...` left `dims`, `plot` and `z` pointing at the old geometry.
- `item.type = EUMType.Wind_speed` on a Water Level item in metres was accepted, although the `unit` setter refuses the same combination from the other side.
- `g.node_coordinates[:,2] = z` (the documented way to change bathymetry) left `element_coordinates`, `boundary_polygons` and the other cached values unchanged, and `g.node_coordinates = arr` accepted an array of the wrong length.
- `ds[0].name = "x"` renamed the item but not the Dataset's key for it, so `ds["x"]` raised `KeyError` while `ds[old_name]` still worked — and the written file used the new name.

These are now properties that validate and invalidate what they invalidate, `node_coordinates`/`codes` are handed out as read-only arrays, and the Dataset derives item names from the DataArrays instead of keeping a second copy. The name-mirroring rewrite also removes the hand-maintained list of reserved attribute names: `__getattr__` is only reached after normal attribute lookup fails, so a real member always wins over an item of the same name.

## Breaking changes

Code that edited these arrays in place must copy, modify and assign back:

```python
nc = geometry.node_coordinates.copy()
nc[:, 2] = new_bathymetry
geometry.node_coordinates = nc
```

- `GeometryFM*.node_coordinates` and `.codes` raise on in-place assignment (`Mesh.zn` and `Mesh.codes` are unaffected and remain the shortest route).
- `ItemInfo.type` rejects a type that the current unit is not valid for.
- `DataArray.time` rejects a time axis that does not match the data.

The [mesh user guide](https://dhi.github.io/mikeio/user-guide/mesh.html) is updated accordingly.

### DIFF
--- a/docs/user-guide/mesh.qmd
+++ b/docs/user-guide/mesh.qmd
@@ -121,15 +121,16 @@ Assume that we want to have a minimum depth of 2 meters and change the open boun
 ```{python}
 g = msh.geometry
 print(f'max z before: {g.node_coordinates[:,2].max()}')
-zc = g.node_coordinates[:,2]
-zc[zc>-2] = -2
-g.node_coordinates[:,2] = zc
+nc = g.node_coordinates.copy()
+nc[nc[:,2]>-2, 2] = -2
+g.node_coordinates = nc
 print(f'max z after: {g.node_coordinates[:,2].max()}')
 ```
 
+Note that `node_coordinates` and `codes` are read-only: modify a copy and assign it back, as above. Otherwise values derived from them (e.g. `element_coordinates`) would be left unchanged.
 
 ```{python}
-c = g.codes
+c = g.codes.copy()
 c[c==2] = 1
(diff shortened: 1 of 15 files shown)


ecomodeller:
@copilot the idea of this PR is sound, review to make sure it doesn't do any harm.

PR #1039, opened 2026-09-01
Global options: item display and progress bars

### LINKED ISSUE #1037: Printing a Dataset from SW point output tells me nothing
I run MIKE 21 SW and extract results at a list of output points. Because dfs0
has no concept of dimensions, every variable/point combination ends up as its
own item, so these files routinely have tens or hundreds of items. There is a
file in your own test data with the same shape as what I get out of SW:
`tests/testdata/sw_points.dfs0`, which has 60 items named like
`Buoy 2: Sign. Wave Height`, `Point 4: Sign. Wave Height`, and so on.

When I read one of these and print it, I get this:

```
>>> ds = mikeio.read("sw_points.dfs0")
>>> ds
<mikeio.Dataset>
title: Point
dims: (time:11)
time: 2017-01-01 00:00:00 - 2017-01-01 10:00:00 (11 records)
number of items: 60
```

So the one thing I actually wanted — which items are in the file, and in which
order — is the one thing it refuses to tell me. I end up typing
`[i.name for i in ds.items]` every single time, which is worse than just
printing the Dataset would have been.

I understand why you don't want to dump 60 lines into a notebook by default.
But pandas and xarray both solved this a long time ago: they show the first few
entries, tell you how many were left out, and let you raise the limit if you
want more. `pd.set_option("display.max_rows", ...)` and
`xr.set_options(display_max_rows=...)`. Could mikeio not have something similar?

It would be nice if this were configurable rather than hardcoded — I would set
it high once at the top of my scripts and forget about it. Right now the cutoff
seems to be baked in at 10 and I have no way to influence it.

ecomodeller:
Closes #1037.

Printing a Dataset or dfs file with many items showed only `number of items: 60`, withholding the item names and their order — the main thing a repr is for. Files of this shape are routine rather than exotic: MIKE 21 SW point output produces one item per variable/point combination, because dfs0 has no concept of dimensions, so tens or hundreds of items is normal. The workaround people fall back on, `[i.name for i in ds.items]`, is worse than printing the object would have been.

So list the first items and summarize the rest, as [pandas](https://pandas.pydata.org/docs/reference/api/pandas.set_option.html) and [xarray](https://docs.xarray.dev/en/stable/generated/xarray.set_options.html) do, and make the cutoff configurable instead of hardcoded at 10:

```python
mikeio.set_options(display_max_items=100)
```

`set_options` also works as a context manager, mirroring [`xarray.set_options`](https://docs.xarray.dev/en/stable/generated/xarray.set_options.html); `display_max_items=None` lists all items, `0` restores the old count-only line. `get_options()` returns the current values.

The new limit governs all six item-listing reprs. They previously disagreed at the boundary — with `< 10` vs `> 10`, a file of exactly 10 items printed only the count from `mikeio.open()` but listed all ten after `.read()` — so this also removes that inconsistency.

Unlike xarray, `set_options` takes explicit keyword arguments rather than `**kwargs`. xarray needs the latter for its ~20 options; with two it would only cost autocompletion and type checking at the call site — a misspelled option should be a `TypeError` from Python, not a `ValueError` from a name registry.

## Progress bars

Having one place for options exposes the second, older mess. Progress bars had three switches:

- `generic.show_progress`, a module global read at call time, defaulting to `True`, absent from `generic.__all__` and from the docs.
- `show_progress = False`, a class attribute on the readers, to be assigned per object — which `mikeio.read()` gives no way to reach, since it constructs the reader itself. So there was no way at all to get a progress bar out of `mikeio.read()`.
- `generic.diff` read neither, so its bar was always on.

All three are replaced by the option:

```python
with mikeio.set_options(show_progress=True):
    mikeio.generic.quantile("large.dfsu", "quantiles.dfsu", q=[0.25, 0.75])
```

Default `False`, so the generic functions no longer draw a bar unasked; reading is unchanged. The global and the class attributes are removed rather than deprecated — neither was documented or tested, and nothing in the repo set them.

dfs3 had no progress reporting at all, despite having the slowest reads; it now uses the same call as dfs1 and dfs2. dfs0 stays without one — it is fast enough that a bar would only add noise.

Documentation: `set_options`/`get_options` are in the API reference under a new *Options* section, the Dataset user guide gains a *Display options* section, and the Generic user guide a *Progress bar* section.


ecomodeller:
@copilot review this PR

